Issue #65, opened 2018-12-07
Find in Article (Alt-F) not working


lizy14:
I'm with latest version from Microsoft Store (0.9.9.81 Beta as shown in About), Windows 10 1809.

I've tried this on multiple articles from multiple ZIMs, but it won't work at all.

![image](https://user-images.githubusercontent.com/9985286/49661912-5f398680-fa85-11e8-8c36-9f3662f5ba3a.png)

Jaifroid:
@lizy14 Thanks for the report, I'll investigate!

Jaifroid:
@lizy14 I can't reproduce this problem on my system, with a fresh install of Kiwix JS from the Store. However, I don't yet have W10 1809, my machine has not yet received the update. I'd be very surprised if something specific in 1809 introduced an incompatibility in EdgeHTML with the JS code that does the in-page search, though anything is possible.

On Monday, I aim to release a major new update to Kiwix JS that supports TEDx video playback and "download" of epub ebooks from Project Gutenberg ZIMs. It doesn't have direct changes to the Ctrl/Alt-F in-page search code, but the update might reset something not working in your install. When you get the new update, let me know if the issue persists.

Alternatively, you could try uninstalling Kiwix JS (right-click the app icon and Uninstall) and then re-installing it from the Store. You won't have to re-download your ZIM files, they won't be touched by the uninstall, but you will have to pick their folder or the specific file again from the Config page. Or wait for the update after Monday, and we'll investigate further if the issue persists.

Jaifroid:
@lizy14 , Kiwix JS version 0.9.9.85 Beta is now available in the Store. Could you please try it and see if your issue is fixed? Please make sure you see the splash screen like below after update (or check the version number in About). If your issue persists, please let me know which ZIM in particular you use, so I can try to reproduce. On the built-in Ray Charles test ZIM, I am not seeing any problem, nor with others I've tested. Many thanks.

![image](https://user-images.githubusercontent.com/4304337/49732281-fd8c3e80-fc75-11e8-9b54-7d6df4b4f6ea.png)

lizy14:
I've updated to 0.9.9.85 Beta, and my issue persists. 
By accident, I realized that it relates to how quickly I type. 

![screenshot](https://user-images.githubusercontent.com/9985286/49747229-eaad5480-fcdd-11e8-996f-19a1d170bc21.gif)

However, for searching Chinese phrases, I still can't get it to work.

[...]

PR #66, opened 2018-12-18
Fix Find in Article for Chinese

lizy14:
#65


Jaifroid:
@lizy14 OK, thanks for the changes, the code looks good, and I confirm it doesn't break searching for accents in Romance languages. There may be some performance issues with searches on very long articles where we end up with a regex like `/expedition|i/` where every single instance of `i` is highlighted, but that's a problem with the original code, not with this PR. It might be better to search for the actual typed string with languages that use spaces, substituting a space with `\s` in the regex rather than with the regex OR character `|`. But that's a different issue. If you're happy with the Chinese-language testing you've been doing, let me know, and I'll merge the PR.

PR #138, opened 2021-02-04
First PWA Dockerfile

### LINKED ISSUE #137: Query about serving code from secure server to enable Service Worker mode in the app
@kelson42 A bit of advice, please. I can now enable Service Worker mode in the UWP app if (inside the app) it changes its address from local code to a secure server capable of serving `https:`. It effectively then runs as a Service-Worker-driven PWA, needing one-time access to the server to cache its own code using the Cache API. It then runs fully offline (apart from self-updating if a new Service worker is released on the server, and if the server is accessible). If running in a UWP shell, it also has access the full WinRT APIs for FS access, etc.

This is a big advance for users of the UWP app, as they can now use the dynamic UI with Gutenberg, TED, PhET, etc. This will be release 1.2 of the app (currently 1.1.4).

To enable this, I am currently serving the code from GitHub:

https://kiwix.github.io/kiwix-js-windows/

(This can also be launched as a PWA taking advantage of the File System Access API in recent Chromium browsers.)

While I would be perfectly happy serving it from there in princple, I understand there are sound reasons for not handing over too much control to one corporation, and further I use that implementation for development, bug-fixing and test releases where I need to test the behaviour of the app when served from a secure domain. So it's not ideal for a stable production app.

What would you suggest here? I see two alternatives:

1. A slot on the Kiwix server, but I'd need credentials / secret. In theory I could create a folder on `download.kiwix.org` with the existing CI secret, but I don't know if it's the appropriate server to serve dynamic app code, and I'd need to agree a folder location with you;
2. I could make a separate "production" folder on the GitHub site, e.g. `https://kiwix.github.io/kiwix-js-windows/pwa/` which would have a stable version of the code in it only updated with an official release. This has the advantage of being easy for me to update. One disadvantage is that it duplicates code on the server.

kelson42:
Fixes #137

### DIFF
--- a/Dockerfile.pwa
+++ b/Dockerfile.pwa
@@ -1,0 +1,11 @@
+FROM nginx:latest
+
+EXPOSE 80
+
+COPY ./manifest.json /usr/share/nginx/html
+COPY ./pwabuilder-sw.js /usr/share/nginx/html
+COPY index.html /usr/share/nginx/html
+COPY CHANGELOG.md /usr/share/nginx/html
+COPY LICENSE /usr/share/nginx/html
+COPY www /usr/share/nginx/html/www
+COPY archives/README.md /usr/share/nginx/html/archives/README.md


Jaifroid (approved):
Thank you very much @kelson42 !

PR #210, opened 2021-11-10
Use node.fs for reading content of directories in Electron apps

### LINKED ISSUE #206: Update Electron builds to use latest stable
The version of Electron used in Electron builds of this app is currently 10.4. The reason I got stuck on this build is that a significant API change was introduced after 10.4 which requires a different way of exposing Electron APIs to code that runs in the browser. Electron has since advanced to 15.2, and GitHub is labelling this repo as having a vulnerability due to using an old Electron.

So it is time to grasp the nettle and update the API usage. This affects principally the use of Node fs, but it also impacts on Service Worker and Cache API. At the same time, we should move away from using the default file: protocol, and register a specific protocol such as app: or electron-app: .

Jaifroid:
This PR fixes #206. It also enhances file picking with Electron. In sum, the PR:

* Updates API usage so that Kiwix JS Electron can run on the latest Electron builds;
* Modernizes the file and directory picking in the Electron app, so that it has equivalent functionality to the UWP, PWA and NWJS apps. In other words, the user can now pick a directory of ZIM archives, and the app will list all of the archives, allowing the user to select amongst them by clicking on each archive name;
* Provides graceful return to Configuration and file selectors if a previously picked archive can no longer be found.

NWJS provides this functionality "natively" via the File System Access API, which it hacks so that the user does not have to give permission for the installed app to access files and directories (after the file or directory has previously been picked). Unfortunately, Electron has so far only partially implemented the File System Access API (see https://github.com/electron/electron/issues/28422) and there is no sign of interest in fixing it. Hence it is necessary to use Electron-specific (Node) APIs to access the FS.

TODO:

* The final thing to do is provide access to split archive parts in a folder, so the user only has to select one part, and the rest are automatically selected (this is supported in the UWP, PWA and NWJS apps).

### DIFF
--- a/main.js
+++ b/main.js
@@ -66,19 +66,22 @@ function createWindow() {
 function registerListeners() {
     ipcMain.on('file-dialog', function (event) {
         dialog.showOpenDialog(mainWindow, {
+            filters: [
+                { name: 'ZIM Archives', extensions: ['zim', 'zimaa'] }
+              ],
             properties: ['openFile']
-        }).then(function (filePaths) {
+        }).then(function ({ filePaths }) {
             if (filePaths.length) {
-                event.reply('fileSelect', filePaths[0]);
+                event.reply('file-dialog', filePaths[0]);
             }
         });
     });
     ipcMain.on('dir-dialog', function (event) {
         dialog.showOpenDialog(mainWindow, {
             properties: ['openDirectory']
-        }).then(function (filePaths) {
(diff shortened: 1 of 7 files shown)


Jaifroid:
Split archives are now supported in folders. Drag-and-drop of split archives is also now re-enabled, but note that the user cannot then re-open the dragged files automatically on next app launch (in the Electron app) because there is no path information provided with a dropped file.

PR #215, opened 2021-11-30
Fix UI bug with secondary displays in UWP app

### LINKED ISSUE #213: BUG: Search Field Too Wide & Hides Icons: Need Style Purge Feature
Installed the MS Store version.
Downloaded one of the R-to-L ZIM files from app's repo.

Went back to app's small default wiki, the search field is so wide, the top_right icons can not be seen.

Long time coder & UI designer.

Tried:
Many re-installs
Many pref changes in settings, plus many combo of
Downloaded several more ZIM files from app repo
etc.

NO GOOD.
Even after reinstall the app seems fine. But when you close & relaunch app, same problem.
Uninstalled (Windows settings REMOVE APP) many times & reinstalled from MS Store, no good, same bug.
Tried the other version from website but totally different and no dark modes nor dark_article mode. Most coders need dark mode.

Files in the old version (install-able) seem to be in appdata roaming from googling. But MS Store version writes to WindowsApps in prog files folder.
Uninstall removes all folders (usually 4). 
Had installed WikiMed & WikiJourney from MS Store too, same bug happens to them. Uninstalled them too, no good for Kiwix re-installs.

Had a look at RegEdit, lots of references. This is a bad practice. I suggest all app makers use the old ini files in their install directories, Registry is a bad idea.

Bug seems to be the odd wiki download set some style settings that persist (probably in Registry, maybe somewhere else).

SOLUTION:
Add New Feature in Preferences/Settings:

PURGE ALL PREVIOUS STYLES & RESTORE DEFAULT STYLE

This will save the developers from not just this bug, but many future bugs, i promise that. So save yourself many future headaches.

QUICK FIX:
Maybe just setup a var to see if style has changed and just reset that. 

default_style = 1

App checks that on launch and if value is 1 then it will not load other styles.
If it is 0 then it will load whatever style the user had set or a ZIM file set.

But you need a procedure to completely flush the style changes.

If i change zoom with CTRL and + or -, then the bug goes away. But it is a pain to do this on each launch.
This bug needs fixing anyway.

bug_kiwix

bug_kiwix2

Jaifroid:
This PR fixes #213 in the UWP app. It is work in progress.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -726,7 +726,7 @@
                                         </div>
                                         <div class="row">
                                             <div class="col-sm-12">
-                                                
+
                                             </div>
                                         </div>
                                         <label class="checkbox" title="Tap three times for dark | auto | light">
@@ -743,6 +743,11 @@
                                             <input type="checkbox" name="cssWikiDarkThemeInvert" id="cssWikiDarkThemeInvertCheck">
                                             <span class="checkmark"></span>
                                             <b>Use simple inversion for dark theme</b> (should work with all archive types)
+                                        </label>
+                                        <label id="resetDisplayOnResize" style="display:none;" class="checkbox" title="This fixes a bug in the UWP app when it is opened on a secondary display or dragged to an external monitor. It resets the display on app launch and when the app is resized. This increases load time, so turn it off if you do not work with a secondary display.">
+                                            <input type="checkbox" name="resetDisplayOnResize" id="resetDisplayOnResizeCheck">
+                                            <span class="checkmark"></span>
+                                            <b>Fix UI bug with secondary displays</b> (increases load and resize time, turn off if not needed)
                                         </label>
                                     </div>
                                 </div>
--- a/www/index.html
+++ b/www/index.html
@@ -726,29 +726,29 @@
                                         </div>
                                         <div class="row">
                                             <div class="col-sm-12">
+                                                <label class="checkbox" title="Tap three times for dark | auto | light">
+                                                    <input type="checkbox" name="cssUIDarkTheme" id="cssUIDarkThemeCheck">
+                                                    <span class="checkmark"></span>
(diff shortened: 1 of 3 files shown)


Jaifroid:
@washere I wonder if you could test this build of the UWP app to see if it fixes the issue? Please uninstall any current version of Kiwix JS (Store app -- you can keep the Electron and other apps, it's just the Store app or winget version that you need to uninstall). Then please download this appxbundle. When downloaded, double-click to launch it, and it should ask you to install.

https://github.com/kiwix/kiwix-js-windows/releases/download/v1.8.1-dev/KiwixWebApp_1.8.1.0_AnyCPU.appxbundle

To test the fix, please turn it on in Configuration. See screenshot below. When you turn it on, the display will immediately reset (which is actually an app reload). This is so that a user can fix the issue straight away (if they can get into Configuration). This is a clunky fix, but it works with my testing. The display will also reset if you resize the app (while this option is turned on). Reset on resize is the only way to deal with the drag-to-monitor issue. I can't think of a better way. NB the displayed option is only available in the UWP app, as this is the only version of the app that is affected.

image

Jaifroid:
@washere There's now a new package:

https://github.com/kiwix/kiwix-js-windows/releases/download/v1.8.1-dev/KiwixWebApp_1.8.11.0_AnyCPU.appxbundle 

This one should be more agile. It resets faster. It should be more reliable at detecting drag to monitor (and vice-versa) and also a bit faster to reset if the window is resized. Please let me know, in due course, if it works for you.

I'm thinking about how to make the page reload faster by caching the CSS along with the page HTML. But that might be a separate issue for next-but-one release.

Jaifroid:
I'm merging this because I'm overdue with a release of the base app, and would like to do it this weekend. However, changes can still be made if necessary and added to the next release (or this one if they make it in time).

PR #216, opened 2021-12-04
Trap mouse forward and back buttons in UWP app

### LINKED ISSUE #214: Mouse Buttons BACK & FORWARD Functions Do Not Work.
This can be either a Feature Request or Bug or both.

A large percentage of users globally now have mouses with 2 buttons on the side for going back or forward.
Most of these never go back to a mouse without, that is why this large percentage keeps growing. 
Also most professionals from coders to writers to people working daily with computers are only using a mouse with Back/Forward buttons.

Because it saves so much time & is convenient & makes them faster & more productive. They are on many models of wired or wireless (BT, WiFi).
So it is only a matter of time before we have this feature implemented.

Currently the buttons do not work on this release from MS Store.
I tried about 6 different mouse models to test. In my experience once this feature is implemented it works with all mouse models on Windows. Android is another matter, varies on models & OSes. But if you like to do it, it will work with all mouse models with Back/Forward buttons on Windows.

I think after you do it, you will love it yourself and wonder why it was not done before because it is so useful and faster than hunting around the screen for the little icons which is still a bit stone age UI/UX.

Of course Back/Forward mouse buttons  work on Wikipedia etc on all browsers & other Wiki reader apps.

Thanks again for the great work.

Jaifroid:
This PR fixes #214. It is only needed in the UWP app because it is handled natively by browsers and by PWAs. Also by Chromium, therefore by Electron and (although I haven't yet tested it) by NWJS.

### DIFF
--- a/KiwixWebApp.jsproj
+++ b/KiwixWebApp.jsproj
@@ -72,7 +72,7 @@
     <PackageCertificateKeyFile>KiwixWebApp_StoreKey.pfx</PackageCertificateKeyFile>
   </PropertyGroup>
   <ItemGroup>
-    <Content Include="archives\wikipedia_en_100_nopic_2021-10.zim" />
+    <Content Include="archives\wikipedia_en_100_nopic_2021-11.zim" />
     <Content Include="images\BadgeLogo.scale-100.png" />
     <Content Include="images\BadgeLogo.scale-125.png" />
     <Content Include="images\BadgeLogo.scale-150.png" />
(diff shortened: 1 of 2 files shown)


Jaifroid:
@washere Please see the test package release for #214 (Mouse back/forward buttons) here:

https://github.com/kiwix/kiwix-js-windows/releases/download/v1.8.12-dev/KiwixWebApp_1.8.12.0_AnyCPU.appxbundle

Please note this test release does not contain the other fix we are testing, as it is built on master and the #215 hasn't been merged yet. Please uninstall the UWP app before installing this version, to be sure you're starting with a clean slate.

I haven't tested this on a hardware mouse. But it works with the X-mouse utility, which simulates the relevant buttons. Please let me know if it works with your hardware mouse.

Jaifroid:
I have to merge this in order to test if it works in Service Worker mode. Changes can still be made.

Jaifroid:
Appears to be working in Service Worker mode.

PR #245, opened 2022-03-24
Trigger webapp restart

rgaudin:
Note: This app's hosting has already been migrated to our new (k8s-based) hosting. We can discuss different ways to offer this feature but as of now, your app is not reloading automatically on image pushes.

Manually trigger a *restart* on Kiwix k8s after a successful image push.
Previous behavior was to let Docker Hub trigger a restart on (former) hosting via a webhook:
- we don't want to rely on Docker Hub anymore
- hosting platform changed

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -36,3 +36,9 @@ jobs:
           restrict-to: kiwix/kiwix-js-windows
           registries: docker.io
           manual-tag: ${{ github.event.inputs.version }}
+      - name: Restart live webapp
+        uses: actions-hub/kubectl@master
+        env:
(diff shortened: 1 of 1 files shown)


kelson42:
@Jaifroid Could we just have the Docker image in the ghcr.io image repository. This is only "private" (for Kiwix) and no need to have it "public".

Jaifroid (approved):
I don't know any details behind the kubectl action, but this looks great, and thanks for the PR @rgaudin. I think this can be merged now, and then I'll live-test it when I'm back in Bogotá and can finish the WikiMed release for this month.

rgaudin:
✅  I've manually pushed both images to ghcr.io (only last version tag and corresponding latest), updated k8s to use them and deleted the *repo* on Docker Hub.

PR #270, opened 2022-06-07
Incorporate zimit header lookup when dirEntry is not found heuristically

Jaifroid:


Jaifroid:
This covers a few cases where the response to a request is not in the ZIM, but there is a Header which tells us that the requested URL has moved. In such cases, we now look up the corresponding H/ namespace entry and read the location.

PR #277, opened 2022-06-14
Fix mobile/tablet back-button

### LINKED ISSUE #50: Programme the Windows Mobile back button so that toolbar back button is not needed
This shouldn't be hard. Some sample code is here: https://github.com/Microsoft/Windows-universal-samples/tree/master/Samples/BackButton . No JavaScript, but the adaptation to JavaScript is just a matter of wrapping the API with standard procedures.

Jaifroid:
Fixes #50 .


Jaifroid:
Tested and working on Windows Mobile. Also on UWP desktop app (but most PCs don't have back button, some tablets do).

PR #305, opened 2022-10-29
Only display download alert when medium is visible

### LINKED ISSUE #75: The app should only offer to download media files when the controls have scrolled into view
Currently the app offers to download a media file if there is one defined anywhere on the page. This is due to the code that was put in place for playing non-supported video files on mobile, but it also gets triggered for any mp4 or webm video in a normal article. This gets annoying!

Jaifroid:
This PR for KJSWL aims to fix #75 by leveraging the lazy load image system for media elements as well as images (JQuery mode only). This is my oldest current issue, so it's well overdue a fix.


Jaifroid:
The PR achieves most of the goal. Rather than showing the alert when the item is in view, it shows the alert when the item is fetched by the lazy load routine. It isn't perfect: if there are several videos in a block, only one of them will be downloadable. But as this is JQuery only, and JQuery is deprecated, I believe the code is sufficient.

PR #323, opened 2022-12-11
Fix libzim reading in Electron and NWJS apps

### LINKED ISSUE #322: WASM fails to read archives in packaged Electron app
As a follow-on from https://github.com/openzim/javascript-libzim/issues/26, the WASM fails to read archives when running inside a packaged Electron app. It gives the error `stream.node.contents.slice is not a function`.

However, if running the app inside Electron in a non-packaged version (e.g. `npm start`) everything works fine with picked archives. This suggests that the issue is not with the framework, but with the way that packaged apps are read when the provider is node.js as opposed to a File object. It may be necessary to convert the node File to a File object. Alternatively, the issue could be caused by the compressed `asar` file system used by the Electron app (although the archive itself is not inside asar).

Jaifroid:
Fixes #322.


Jaifroid:
Reading archives with NODEFS now works in Electron, which means we can read the FT-index in a packaged app without asking the user to pick the packaged archive.

In NWJS, NODEFS isn't working, apparently due to some bug in enabling Node in Web Worker context. However, as we don't currently provide packaged apps in NWJS format, it will be best to disable NODEFS in the NWJS app for now and open a new issue for that.

Issue #337, opened 2022-12-23
Failed to execute 'readAsArrayBuffer' on 'FileReaderSync' error in Windows Electron App for very large ZIM archives


Jaifroid:
The libzim WASM port is randomly evincing an exception when first attempting to load the full English Wikipedia ZIM in the Windows Electron app (but not the Linux Electron app), as below. Sometimes it loads correctly, and at other times it fails with this exception. **_When it fails, full-text search is unavailable._** This doesn't occur with the full Spanish Wikipedia, so it could be related to the size of the archive. The fact that it sometimes loads correctly suggests a race condition. The fact that it doesn't occur in the PWA version of the app suggests that it affects the NODEFS file system code in the WASM, but not the WORKERFS file system. However, `FileReaderSync` is a method of the standard File API that can be used in Worker context, and is not a Node-specific method.

```
Uncaught (in promise) DOMException: Failed to execute 'readAsArrayBuffer' on 'FileReaderSync':
The requested file could not be read, typically due to permission problems that have occurred after
a reference to a file was acquired.
    at Object.read (file:///C:/Users/xxxxx/AppData/Local/Programs/Kiwix%20JS%20Electron/resources/app.asar/www/js/lib/libzim-wasm.js:1:41693)
    at Object.read (file:///C:/Users/xxxxx/AppData/Local/Programs/Kiwix%20JS%20Electron/resources/app.asar/www/js/lib/libzim-wasm.js:1:57763)
    at doReadv (file:///C:/Users/xxxxx/AppData/Local/Programs/Kiwix%20JS%20Electron/resources/app.asar/www/js/lib/libzim-wasm.js:1:115113)
    at _fd_pread (file:///C:/Users/xxxxx/AppData/Local/Programs/Kiwix%20JS%20Electron/resources/app.asar/www/js/lib/libzim-wasm.js:1:115513)
    at file:///C:/Users/xxxxx/AppData/Local/Programs/Kiwix%20JS%20Electron/resources/app.asar/www/js/lib/libzim-wasm.wasm:wasm-function[600]:0x2dd27
    at file:///C:/Users/xxxxx/AppData/Local/Programs/Kiwix%20JS%20Electron/resources/app.asar/www/js/lib/libzim-wasm.wasm:wasm-function[431]:0x1fda2
[...]

Issue #334, opened 2022-12-20
Electron app has to perform self-reload in order to activate Service Worker


Jaifroid:
When the Electron app starts up, the Service Worker is in a strange state: it is active but "not controlling the current page". To fix this, the app performs a self-reload, which successfully attaches the Service Worker. However, there should be a better way of ensuring SW can control the page. This doesn't occur in the PWA version. It may be caused by the patched URL schema in Electron apps (they use the `file://` protocol, but patch it so that it is recognized as a secure origin).

Jaifroid:
Fixed in #338.

Jaifroid:
The solution to #342 showed that the reload is still happening to activate the Service Worker, albeit very fast and early in the load process. It remains to be seen if this means #337 is not properly fixed.

Jaifroid:
This was reopened, then fixed, but forgot to close.

PR #338, opened 2022-12-24
Fix libzim loading error in electron app

Jaifroid:
This is for fixing #337 and #334.


Jaifroid:
Tested and working in Windows and Linux.

PR #344, opened 2023-01-10
Fix issues with Electron archive and last article load

### LINKED ISSUE #342: WikiMed Electron app not remembering last-visited page (again)
Sigh. I seem to go round in circles with this one. At least it's not a critical feature.

Jaifroid:
Found multiple issues with Electron archive loading in rare situations such as more than one archive in the packaged archive directory (which should be supported). This PR should also fix #342 (not yet).


Jaifroid:
When Electron reactivates a service worker, it appears to be left in an anomalous situation whereby it is active but not controlling the page. We can't do self.clients.claim() because we are not inside the install event, so the only solution is to do a page reload, which successfully reactivates the SW. However, the app then detects that the last page load failed, and prevents going back to the last-visited page (in case that page causes a boot loop).

I've put in code to manage this situation, but I'm not really happy about the app having to reload itself in order to reactivate the SW. It only happens with Electron, so it's probably an Electron bug. It causes some screen flashing on reload of the app.

PR #366, opened 2023-02-13
Provide indication of download progress in Electron

Jaifroid:
This is while downloading archives from the in-app library, or indeed when downloading any item using standard browser download if this is initiated in the Electron app.


Jaifroid:
Currently a blank window opens to deal with the download. Perhaps we can get rid of this window if in an Electron or NWJS app, as it is likely that download can be evoked directly rather than needing to open a new window (this was done to deal with more robust download if done via the system browser when using apps such as the UWP app).

PR #383, opened 2023-03-12
Add missing full text search cancellation

danielzgtg:
This is a port of https://github.com/kiwix/kiwix-js/pull/978 . While the original bug isn't as severe in kiwix-js-windows, nothing broke fixing this. It was thought that this would be nice to port over as it makes the state machine cleaner.


Jaifroid:
Thanks very much @danielzgtg. Tested and working. I'll merge.

Issue #389, opened 2023-04-21
Non-HTML article types (principally in Nautilus ZIMs) are offered for download instead of display


Jaifroid:
This particularly affects "articles" which are images. The app offers to download them. Although they can of course be opened instead of downloaded, common file types such as PNG, GIF and JPEG (as "articles") such as are offered in Nautilus ZIMs, can perfectly well be displayed in the app's iframe or in a new tab.

Issue #388, opened 2023-04-21
Toolbar "scroll-away" feature prevents display of videos in Nautilus ZIMs


Jaifroid:
As per title. This is related to #186. The displacement of the document interferes with the display of the video modal in these ZIMs. Turning off the scroll-away feature is a workaround, but I have a better solution (PR forthcoming).

PR #390, opened 2023-04-21
Fix UI issues with toolbars and downloads

### LINKED ISSUE #186: The click area for PhET expirements is displaced
PhET experiments run in KJSW in SW mode, but the click area is displaced by several pixels. This must be to do with the code for hiding and showing the toolbars, which has some kludges in it in an attempt to work around bootsrap issues. Since no user has complained, I assume very few users attempt to run PhET on the PWA version, but it should be fixed in a generic way.

Jaifroid:
Fixes #186, #388, #389.


Jaifroid:
PR needs thoroughly testing for regressions, especially with Zimit archives.

Jaifroid:
Carefully tested in IE11, Firefox, Chrome. A great improvement.

PR #393, opened 2023-04-30
Remove RequireJS and use native ES6 modules

Jaifroid:
This is a fundamental step towards using modern dependency and module-management tools.

It was necessary to modularize xzdec and zstdec, otherwise they don't use the correct export format for native modules. I had to implement the fallback from wasm to asm using dynamic imports, which may not be functional with older browsers, even when transpiled. We'll have to see what happens with babel or other transpilers.

But it's all working with native ES6 modules.

The [5to6-codemod project](https://github.com/5to6/5to6-codemod) was very helpful in rewriting all the `define([], function () {});` statements as `import x from 'path/file.js';` equivalents, and rewriting the exports.


Jaifroid:
PR now builds a bundled app using Rollup. See https://github.com/kiwix/kiwix-js/issues/554#issuecomment-1532579437 for full explanation.

Note that the app still works in modern browsers without any bundling, using native ES6 modules. When it is bundled and polyfilled it also works in legacy browsers back to IE11.

Next step is to enable libzim WASM loading outside the bundle (as it is best not to bundle binaries, due to massively increased bundle sizes). The other WASMs are already enabled.

Jaifroid:
Libzim WASM loading working fine. I experimented with loading CSS dependencies (principally Bootstrap) using native CSS modules, but there is very patchy support for it via some largely unmaintained plugins, and they don't seem to support native loading of CSS from JS via `document.adoptedStyleSheets`. As we are using an old version of Bootstrap here, there is no need to have it as an NPM dependency. It's more trouble than it's worth.

PR #394, opened 2023-05-06
Add a Promise Queue to prevent overlapping systemAlerts

Jaifroid:
This code should be added upstream as well in Kiwix JS. It will be properly deployed here when we remove RequireJS #393.


Jaifroid:
See https://github.com/kiwix/kiwix-js/issues/841.

PR #422, opened 2023-07-09
Support the `HTMLInputElement:webkitdirectory` API

### LINKED ISSUE #421: Add support for HTMLInputElement:webkitdirectory API (principally for Firefox)
See https://github.com/kiwix/kiwix-js/issues/1032 for the upstream version of this. It is much easier to add this here, since we already support the File System Access API in KJSWL, and the `webkitdirectory` API is a "poormans" version of that API. It is mostly a case of inserting a few hooks, and dealing with loading split archives from any picked directory.

Once the directory is picked, the API is indistinguishable from the File System Access API. The main difference is having to pick the directory of ZIM archives at the start of each session.

Jaifroid:
Fixes #421.


Jaifroid:
Tested and working well on Firefox and Edge Legacy.

Doesn't interfere with IE11 legacy file picking, and doesn't prevent use of newer APIs if available.

Need to check interaction with Electron and NWJS.

TODO: put repeated code into a function.

Jaifroid:
It seems the API has poor support on Android even though `webkitdirectory` is reported as available.

Jaifroid:
No support on any mobile browser: https://caniuse.com/input-file-directory.

Jaifroid:
Prevented false-positive detection on iOS and Android.

Jaifroid:
I've now made a single function to process a directory of files, whether it is provided by `webkitdirectory` or by File System Access API.

It would be nice if the list of files were retained and displayed on reload of the app. Assuming this is possible (it's just a cookie list), then clidking on a file in the list should open the directory picker in Firefox. This would put us at near parity between the two APIs.

Jaifroid:
Last thing to do is make it so refresh button opens directory dialogue.

Jaifroid:
Checks:

- [ ] IE11
- [ ] Edge Legacy
- [ ] UWP
- [ ] Firefox 45
- [ ] Chrome 50
- [ ] Modern Firefox
- [ ] Modern Chromium (Edge)
- [ ] Electron on Windows
- [ ] Electron on Linux
- [ ] NWJS for Windows XP
- [ ] Firefox OS

Jaifroid:
All manual test passed. Ready to merge.

PR #430, opened 2023-07-19
Enable file handling with Electron app

### LINKED ISSUE #426: Add file associations to Electron app
Currently only the PWA and the UWP (legacy) apps can register as file handlers.

However, this is possible with the Electron app adding something like the following to `package.json` (only for the base app, not for WikiMed or Wikivoyage):

```
"fileAssociations": [{
     "ext": "zim",
     "name": "ZIM Archive",
     "role": "Editor"
   },
   {
     "ext": "zimaa",
     "name": "ZIM Archvie",
     "role": "Editor"
   }
]
```

See https://www.electron.build/generated/platformspecificbuildoptions for more info.

Jaifroid:
Fixes #426 at least on Windows. Needs to be tested on Linux.


Jaifroid:
Tested on Linux, and unfortunately it doesn't currently work with the package.json keys used for this in Windows. It may be necessary to register a MIME type instead of a file extension?

Jaifroid:
File association for Linux fixed with https://github.com/kiwix/kiwix-js-windows/pull/430/commits/b2596b214dc19b81d4e036f94cd04929472e1561.

Also fixed #431.

PR #454, opened 2023-09-10
Remove webpHandler if user has selected local styles

### LINKED ISSUE #424: Remove implementation of webpMachine inside ZIM, as it conflicts with the reader-provided version
When testing in Edge 18 (which should support webP, but doesn't in the version on BrowserStack), I noticed that even if I disable our implementation of webpMachine, the version in the ZIM gets loaded, and it seems to take precedence over our implementation in all cases (whether our implementation is enabled or disabled). This is problematic, because the version in the ZIM doesn't handle reading images one-by-one on image-heavy landing pages, which is required to decode them all. As a result, only one or two images can be displayed on these problematic image-only landing pages.

This probably affects Kiwix JS as well.

It needs more testing, and if confirmed, an issue needs to be opened on mwOffliner.

Jaifroid:
Fixes #424, and should make page loads lighter (e.g. for Android). If user turns off the cssCache option, then this patch won't be applied. Perhaps need to make it contingent on manipulateImages being turned on too?


Jaifroid:
Tested and working on BrowserStack.

PR #458, opened 2023-09-17
Enable app as extension for testing only

Jaifroid:
This is only for testing API support in extension context.


Jaifroid:
Code is too much out of date for this to be useful.

PR #459, opened 2023-09-17
Add experimental support for the Origin Private File System

### LINKED ISSUE #349: [PWA] Support Origin Private File System for opening of archives without file picking
The [Origin Private File System (OPFS)](https://developer.mozilla.org/en-US/docs/Web/API/File_System_Access_API#origin_private_file_system) is supported in the latest Chromium Browsers ([including on Android](https://developer.chrome.com/blog/chrome-109-beta/#origin-private-file-system-opfs-on-android)), and I believe there are plans for Firefox to support it as well (the only part of the File System Access API that has any support from Mozilla). This would work a little bit like app-specific storage on Android. Once a ZIM archive is in the OPFS, it should be possible for the app to load it without user prompts or file picking, for a native-like app experience with the PWA. Of course, once in the OPFS, an archive is no longer available to other Kiwix apps, which is precisely the limitation of the Android app currently if installed from the Play Store. This would be most useful for a PWA "packaged app" of WikiMed, for example.

The proof will be in the pudding. It remains to be seen if it is possible to copy very large ZIM archives into the OPFS, or whether such archives are held in memory before being committed to storage. It appears that it is possible to copy in an ArrayBuffer, which is usually a pointer to data on disk, but a lot would depend on the precise transfer mechanism. It should be OK with archives that are less than a couple of gigabytes, but I can imagine hitting some hard limits if trying to copy in 92GB (even if the user had the disk space).

It may be possible to download archives directly into the OPFS, but it would be important also to support a copying-in mechanism, so that users with existing archives are not forced to download them again. Ideally we would also want a copying-out mechanism, which would probably work rather like "downloading" an archive (but in fact would be "downloading" it from the app's OPFS to external storage).

Jaifroid:
Fixes #349 experimentally. The idea is to experiment with the capabilities with the eventual aim of backporting upstream.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -795,7 +795,7 @@
                                         </div>
                                         <div class="col-xs-6">
                                             <div id="OPFSQuota"  style="display: none;"></div>
-                                            <input type="button" class="btn btn-primary btn-inline" value="Select folder" id="archiveFiles" />
+                                            <input type="button" class="btn btn-primary btn-inline" value="Select folder" title="Select a folder containing several ZIM files, and/or all the parts of a split ZIM file." id="archiveFiles" />
                                         </div>
                                     </div>
                                 </div>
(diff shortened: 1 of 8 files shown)


Jaifroid:
Although archives can be picked from the options list on mobile, it is fiddly, and often results in the previously selected archive being loaded anyway. This may have to be fixed simply by offering a simple list of archive links, although presentation would be slightly degraded on desktop.

Jaifroid:
Some things needed to complete this PR:

- [ ] Need to sync OPFS before selecting file to delete or export
- [ ] Test or disable in Electron and NWJS
- [ ] Test that it's already disabled in UWP app
- [ ] ...

PR #494, opened 2023-11-14
Provide backdoor to use libzim for reading ZIM

Jaifroid:
Developer has to open DEV tools after ZIM is loaded and in console write `params.useLibzim = true;`. To prevent any rewriting of HTML also set `params.pureMode = true;` Otherwise some content will be hybrid: main page request read by libzim, but custom backend used for lazy-loading images, and other transformations.


Jaifroid:
Now I've provided a UI to turn it on and off.

Need to check if pureMode is still needed, as it seems to be bypassing rewrites and style transformations.

PR #499, opened 2023-12-06
Update bundling method for jQuery

### LINKED ISSUE #492: Port the solution adopted for loading jQuery and Bootstrap from Kiwix JS
Although rollup builds the bundle correctly, and it works, we're using an outmoded way of including jQuery. It effectively tries to minify the already minified jQuery and Bootstrap bundles. It would be much better to adopt the more robust solution I developed for Kiwix JS, which is to load these dependencies from their npm packages (unminified), and pass the unminified versions through the minifier. This should eliminate the current build warnings shown below.

image

Jaifroid:
Fixes #492, but only for jQuery because we are using an old bootstrap here, and it will be better to do that work after upgrading or changing the UI framework.

Remaining to be done:

- [x] Fix the GitHub Pages implementation so that the module files are exposed
- [ ] ~Fix the dockerfile so new JQuery is included in pwa.kiwix.org~ (this isn't necessary, because only the bundled version is uploaded to pwa.kiwix.org)


Jaifroid:
Appears to be working as intended.

PR #500, opened 2023-12-06
Add replay system capability

Jaifroid:
Port Replay-based Zimit reading to PWA.


Jaifroid:
Todo:

- [ ] Fix JQuery support for Zimit (was working, now broken with replay system incorporated)
- [ ] Fix recognising external Replay links and PDFs on click
- [ ] Add a way to use legacy Zimit support (developer option for testing fallback)

Jaifroid:
Darkreader support is a bit hit-and-miss, because it is currently added after the page has loaded. But for a preview, this will do.

PR #508, opened 2023-12-15
Stabilize DarkReader with Zimit ZIMs

### LINKED ISSUE #507: Stabilize DarkReader with Zimit ZIMs
We currently have issues (with Zimit ZIMs read by Replay and subsequently transformed by DarkReader) such as some flash of incorrect theme, banner images not reloaded correctly by DarkReader due to the iframe being hidden at the point of rewrite, or (rarely) the iframe not being unhidden leaving the user looking at a blank screen.

All these are fixable.

Jaifroid:
Fixes #507.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,21 @@
 # Changelog
 
-## In-progress release 2.8.6
+## In-progress release 2.8.7
+
+FIXES since 2.8.5 was announced:
+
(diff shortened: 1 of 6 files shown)


Jaifroid:
Tested on desktop and mobile (Android), but still issues on Firefox, probably because the onload event fires too soon on Firefox.

Jaifroid:
But the Firefox issues are mostly confined to one site (practicalaction), and I don't think there's any more we can do.

PR #511, opened 2023-12-19
Attempt to avoid sandbox for PDFs on Android with Zimit archives

### LINKED ISSUE #504: On Android, PDFs are blocked from opening in Chromium because a new tab doesn't escape the sandbox!
On ***every other OS*** (as far as I can tell), opening a new tab escapes the sandbox of the iframe. On Android, however, this isn't the case, as the new tab is opened within the scope of the PWA, rather than in the browser.

Either we have to force a download on Android (at least for Chromium, Firefox doesn't see PDFs as a security threat), or else there might be some other workaround to force the link to open in a browser rather than in the scope of the PWA...

Needs more investigation.

Jaifroid:
Fixes #504. Needs testing carefully.


Jaifroid:
OK, finally seems to be working on Android, also with "PDF" links that are in fact HTML redirects (such as those at readygov).

Issue #521, opened 2023-12-29
Opening a Zimit article in a new window (in the PWA) always opens a new browser tab instead


Jaifroid:
Despite the value of the "open in new window instead" setting, this appears to be ignored for Zimit-based archives.

Jaifroid:
Fixed as far as possible with #522.

PR #522, opened 2023-12-29
Reset pointers to articleContainer on new ZIM load

### LINKED ISSUE #520: After opening article in new window, opening any new ZIM fails to display landing page
It appears that the target container for loading the content isn't reset to the iframe (which is the only place a new ZIM's content can be loaded). It's probably a simple fix, but needs debugging.

Jaifroid:
Fixes #520. Needs testing in app context, but we might as well include #521 too.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -4009,6 +4009,10 @@ function archiveReadyCallback (archive) {
     appstate.selectedArchive = archive;
     // A blob cache significantly speeds up the loading of CSS files
     appstate.selectedArchive.cssBlobCache = new Map();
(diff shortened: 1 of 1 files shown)


Jaifroid:
#521 Is tricky to fix, and I've done what I can. What appears to happen now is that the article will open in a new window (if option is turned on) the first time that an article is ctrl-clicked, but while that window is open, on subsequent clicks on different articles, the browser will ignore requests to open more windows, at least until the initial one is closed. This could be an anti-popup measure, though I hadn't seen it before.

PR #525, opened 2024-01-01
Increase timeouts waiting for Service Worker

Jaifroid:
With the addition of replayWorker, which is quite heavy to cache, the Service Worker may not have fully loaded before the app decides it has to issue a refresh (an essential workaround for stuck Service Workers). Because a refresh dismisses the initial welcome dialogue, this can be problematic for first install.


Jaifroid:
Once we have good values, this will need to be ported upstream.

Issue #536, opened 2024-01-22
Add support for Zimit v2.0 archive types


Jaifroid:
There are a few parts to this:

- [x] #537
- [x] #538
- [x] #541
- [x] #542
- [x] #544

Further info on the second item: as these ZIM types are not currently detected as zimit2, the existing handler thinks that all links in a zimit2 archive are external (due to Wombat rewriting). We either have to disable wombat rewriting for clicks on links as was done in https://github.com/kiwix/libkiwix/pull/1036, or else we have to use our existing detection method for Zimit files minus the header lookup. I'm not sure which will be better just now.

PR #537, opened 2024-01-22
Add a way to detect that an archive is zimit2

Jaifroid:
Fixes one of the bullet points in #536.


Jaifroid:
It works like this:

1. Assume all archives are 'open' as a baseline
2. If we find warc-headers, it must be zimit classic (this is true currently, but might not be true in the future)
3. If we don't find warc-headers, but the archive was scraped by warc2zim, then it must be zimit2

PR #538, opened 2024-01-23
Fix zimit2 external links and PDFs

Jaifroid:
Fixes second bullet point of #536.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -4042,7 +4042,7 @@ function archiveReadyCallback (archive) {
       archive.zimType !== 'open') {
         if (params.imageDisplay) params.imageDisplayMode = 'all';
         if (params.zimType !== 'zimit') {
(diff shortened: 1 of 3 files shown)


Jaifroid:
Tested (and working`*`) on the following zimit2 archives in **ServiceWorker mode**:

* 100rabbits_en_2024-01.zim
* mes-quartiers-chinois_fr_all_2024-01.zim
* solar.lowtechmagazine.com_en_all_2024-01.zim

`*` There is a small issue with the currently loaded page in the iframe being blanked out if the user opens a link in a new tab. The link in the new tab works fine and is navigable, but the iframe is put out of commission by this (and it is hard to open the landing page in the iframe, it keeps redirecting to the tab). This needs investigating.

**JQuery mode** has odd effects on mes-quartiers-chinois: images are all detected as hidden by default, even though the UI option to hide images is not turned on. Also needs investigating.

Jaifroid:
PDFs are working with zimit2 archives, but now this code breaks PDFs in classic zimit archives! 😡

PR #543, opened 2024-01-26
Correct detection and use of darkReader

### LINKED ISSUE #541: DarkReader is non-functional with zimit2 archives
It simply appears to be a problem of not injecting the DarkReader script correctly somewhere in the app's code. Needs investigating.

Jaifroid:
Fixes #541


Jaifroid:
There is flash of incorrect theme. Need to delay unhiding of iframe.

Jaifroid:
Now working, though there is the occasional FOIS with zimit classic on some archives. I think it's very minor.

Main thing to work on now is jQuery mode, pretty much broken for zimit2, though working for zimit classic, but that is a separate issue.

Issue #542, opened 2024-01-24
Blue placeholders display instead of images when opening a zimit2 archives in jQuery mode


Jaifroid:
This is a bit bizarre, and it doesn't occur upstream. Tested on 100rabbits and mesquartierschinois, so I suspect it is all such archives. This is a problem with the PWA, not with the zimit2 format. The blue placeholders are the same ones produced if the user requests manual extraction of images, and indeed they can be clicked and images are then manually extracted (one-by-one).

It's not clear why hiding of images is activated only for zimit2 archives in jQuery mode. Needs debugging.

Jaifroid:
While the blue placeholder issue seems to be fixed, we now have no images at all with Zimit 2 in jQuery mode.

PR #545, opened 2024-01-29
Fixes for zimit2 in jQuery mode

### LINKED ISSUE #544: Fix safe-mode (jQuery mode) reading of zimit2 archives
Apart from #542, safe-mode reading of zimit2 files is pretty much broken now.

It should be very similar to opnZIM jQuery, but something is interfering.

Jaifroid:
Fixes #544 and #542


Jaifroid:
Probably as far as I can go now. We're still missing (but have always missed) background images in CSS. It's just that some Zimit ZIMs use them extensively.

PR #550, opened 2024-01-31
Fix UWP handling of untransformed Wombat links

Jaifroid:
Works around relative links.


Jaifroid:
It looks like I can't go any further with this. Videos, unfortunately, fail without wombat.js, so I've opened #551.

Issue #514, opened 2023-12-21
Flash of incorrect style when loading a Zimit ZIM for first time with DarkReader or when using the random button


Jaifroid:
While I have eliminated FOIS when using links within a loaded ZIM that has DarkReader enabled, incorrect style is still showing in the following circumstances:

* Loading a ZIM for the first time, or pressing the Home button
* Using the random button

These are minor issues, and it is a matter of porting the code that waits for DarkReader to have installed itself before showing the contents of the iframe, but it should be done for aesthetic reasons (and, marginally as an a11y issue).

Jaifroid:
FOIS now affects most Zimit v2 ZIMs with DarkReader. The reason is because these ZIMs are faster to read (due to the pre-processing). It is difficult to detect when DarkReader has completed its transformations (this is some time after the `enabled` flag is set), so we have to introduce more time-based delay.

PR #559, opened 2024-02-02
Reduce flash of incorrect style

Jaifroid:
I find FOIS to be quite annoying, and it's potentially an accessibility issue too.

This PR is to find the correct timings for most situations.


Jaifroid:
Reduces the problem, though it's not a total fix for #514.

Issue #551, opened 2024-01-31
With zimit2 in the UWP app, provide fuzzy search workaround for YouTube videos due to incompatibility with wombat.js


Jaifroid:
Unfortunately, due to the incompatibility of wombat.js with the UWP app, we can't use the native support for finding the video BLOB in zimit2 archives.

Instead, we'll have to do adapt the fuzzy search as implemented for zimit1 legacy mode. Unless https://github.com/openzim/warc2zim/issues/171 can target Edge 18, which seems very unlikely.

This hack would take some time, so I won't implement it for v3.0, especially as zimit2 is not yet officially released.

PR #561, opened 2024-02-04
Fix for legacy video player

### LINKED ISSUE #512: Zimit videos that work fine in desktop browser fail to work on Android
There is no difference in the underlying code, and this isn't the libzim issue with video on Android, because we're not using libzim for this extraction. So it's a little strange. I suspect a resolution or a codec issue, if the dynamic player is requesting a specific version for Android.

### LINKED ISSUE #560: Video broken with legacy zimit1 playback method
A change in the way replay links are handled in the Service Worker broke the legacy zimit1 reader's video resolver (for browsers that don't support replayWorker). It's to do with not being able to clone a request that relates to streaming content.

Jaifroid:
Fixes #560. Also will try to fix #512 and #551 if possible.

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -626,11 +626,16 @@ function zimitResolver (event, rqUrl) {
         // If the reqUrl is not the same as event.request.url, we need to modify the request
         // Note that we can't clone requests with streaming bodies, hence we check for bodyUsed (see https://developer.mozilla.org/en-US/docs/Web/API/Request/clone)
         var rtnRequest;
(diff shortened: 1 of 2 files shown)


Jaifroid:
We can't fix #512 in this PR, because it is a problem with the fuzzy substitutions in zimit2 software. See https://github.com/openzim/warc2zim/issues/144. As for #551, it's not so urgent and it's best to wait for the format to settle, because there's some intensive work being done on zimit2 video right now.

Jaifroid:
Fixed as far as we can get for now.

PR #565, opened 2024-02-08
Properly fix legacy zimit video

### LINKED ISSUE #564: Zimit1 video broken in legacy UWP app
Although I fixed the legacy playback method in Chromium browsers in #560, it seems the issue persists for the Edge HTML legacy webview, whereas video playback definitely used to work in this context. To be clear, this affects Zimit Classic, and is a separate issue from #551 (which affects Zimit2), though the remedy might be somewhat similar.

It is bizarre that the previous fix works in Chromium browsers set to use the legacy playback method, but not in the UWP app. Neither browser uses Wombat in this configuration.

Sigh.

Jaifroid:
Fixes #564, hopefully.


Jaifroid:
To be clear, the fix only applies to YouTube video in ServiceWorker mode. Video in JQuery mode is mostly unsupported, and I'm not sure it's worth supporting even though it is technically possible using the HTML5 video workaround.

PR #577, opened 2024-03-14
Workaround for HTML URLs with percent-encoded querystring separators in zimit2

Jaifroid:
We need to fix clicking on a link which has querystring separators like `?` and `=` encoded. We find these in zimit2 at least because they have been introduced as a workaround for the fact the Kiwix Serve can't handle querystrings. See https://github.com/openzim/libzim/issues/865 for details.


Jaifroid:
Small adjustments in the light of https://github.com/openzim/warc2zim/pull/218

PR #580, opened 2024-03-26
Fix failure to print HTML books in Gutenberg ZIMs

### LINKED ISSUE #579: Unable to print the HTML version of Gutenberg books in SW mode
When printing, the page is reloaded, resulting in the app reloading the landing page and not the opened book. Workaround is to Ctrl-click the HTML icon form the cover page and print from the opened tab. However, there are overlays on the page which would need to be filtered out.

Additionally, opening a Gutenberg HTML e-book in JQuery mode results in images being set to manual mode (you have to click them to display them).

Jaifroid:
Fixes #579.

So far, it can now print the HTML in the iframe, but images are not being retrieved.

Also fixes the bug with mobile vs desktop printing.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -463,10 +463,12 @@ function printIntercept () {
     if (!innerDoc) {
         return uiUtil.systemAlert('Sorry, we could not find a document to print! Please load one first.', 'Warning');
     }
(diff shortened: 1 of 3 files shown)


Jaifroid:
Now working in SW mode and jQuery mode. Images fixed in jQuery mode.

PR #581, opened 2024-03-31
With FSA persistent permissions go straight to article view

### LINKED ISSUE #573: When using persistent FSA permissions, go straight to article view without showing Config first
When using persistent permissions with the PWA, there is a slightly clunky flow. Instead of opening straight to the article view, like in the Electron app, the PWA goes first to Configuration and then switches to the article view. This is probably the result of some prior code that switches to Configuration view if the browser is using the File System Access API. However, this should be done only if permission cannot be obtained.

Note that this is already the case if the user is using the OPFS.

Jaifroid:
Fixes #573. Need to check for regressions when FSA permissions not granted.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1119,7 +1119,7 @@ function getNativeFSHandle (callback) {
                     if (callback) {
                         callback(handle);
                     } else {
-                        searchForArchivesInPreferencesOrStorage(true);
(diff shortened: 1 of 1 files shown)


Jaifroid:
Tested on Android, Edge PC, Firefox PC. All seems good.

PR #595, opened 2024-04-30
Add Wikipedia preview function

### LINKED ISSUE #278: Support "page preview" feature of Wikipedia and Wikivoyage
This is the same as upstream https://github.com/kiwix/kiwix-js/issues/719. I would be happy to work on this upstream, but it might be better to experiment with the feasibility of this feature here, only because I already have the click infrastructure set up, whereas https://github.com/kiwix/kiwix-js/pull/680 failed. Happy to be corrected if this supposition is wrong.

Having said this, the modern way to do this would be to attach the required DOM nodes programmatically *only* when the user hovers over (or long-presses) a hyperlink. Detection would be done at iframe (or tab/window) level, making this compatible with SW mode -- this is similar to the mechanism used in https://github.com/kiwix/kiwix-js/pull/862.

This ticket is medium-to-long term. I don't have time to work on it right now, but it would be cool. Expected end result is as below. Our simple page extraction is very fast (a few tenths of a second on a good machine and modern browser), and we wouldn't be extracting assets other than one image (also very fast to extract).

image

Jaifroid:
Fixes #278.

This is currently specific to Wikimedia ZIMs (i.e. those scraped by mwOffliner) because the HTML has to be parsed based on certain tags provided by mwOffliner.

@kelson42 has suggested that a more generic way could be to snapshot a rendered (but hidden) view of the article and display a png. This would be more generic, although we need to think whether users would expect or even appreciate a generic preview. It's certainly the case that Wikipedia users expect such a feature (and this is the easiest case to cater for).

This feature should be ported upstream to Kiwix JS, but currently the click "infrastructure" is missing there in ServiceWorker mode (or rather, it is done differently), so the code would need to be adapted. See https://github.com/kiwix/kiwix-js/issues/719.


Jaifroid:
Some to-do items:

- [ ] Fix dark theme support (background is currently transparent with dark theme which produces illegible mix of foreground and background
- [ ] Fix calculation of base URI: i.e., get absolute ZIM url for images (otherwise they will not load automatically in SW mode)
- [ ] Consider supporting Restricted mode, i.e. manually extracting the image from the ZIM
- [ ] Add arrow pointer in direction of text (ideally situated at site of mouse pointer)
- [ ] Add some options to the box: principally, open in new tab or window
- [ ] Make box show with long press (supporting touchscreen)

Jaifroid:
Feature is tested and working on Chromium and Firefox for desktop: mouseover is best method, but long-press also works for touchscreen users, and keyboard (tabbing into links) also evokes the popover. Also tested on UWP app, all working fine with mouseover and touch.

On Android, an installed PWA can use the feature with long-press, and I've included different sizes for portrait and landscape orientation when there is a narrow screen. However, in Firefox for Android long-press seems to be hard-wired to the context menu and is not overridden. However, Firefox for Android cannot load large ZIM archives due to https://bugzilla.mozilla.org/show_bug.cgi?id=1807360, so I doubt anyone uses the PWA in that browser on Android.

I have yet to test Safari / iOS in simulator.

Jaifroid:
OK, iOS at least on an iPad has its own "hard-wired" preview feature which my code doesn't interfere with, so I think that's acceptable. See screenshot.

image

Jaifroid:
Feature is working great on Safari desktop on macOS. Both mouseover and keyboard (macOS doesn't have touch).

image

Jaifroid:
At least on Android, the icons top right are not consistently attached.

Jaifroid:
Now working in Firefox for Android (with long-press), but close icons are not registering, despite some tweaks. Needs debugging.

Jaifroid:
Finally icons are working on Firefox!

Jaifroid:
All tested and working in Chrome (>=49), Firefox (>=68 - earlier versions work but don't support WebP), UWP (Edge Legacy), Safari on macOS. Ready to merge.

PR #604, opened 2024-05-16
Add update progress and do not show GitHub updates if PWA update is needed

### LINKED ISSUE #591: If an Electron app is doing an auto-update, remove the notice to go to GitHub for the latest release
There is potential for confusion, and even the possibility that the user will uselessly download the update, only to find it has auto-updated. I should replace the notice with something like "Download in progress...".

Jaifroid:
Fixes #591

### DIFF
--- a/main.cjs
+++ b/main.cjs
@@ -228,7 +228,7 @@ app.whenReady().then(() => {
         mainWindow.webContents.send('update-available', info);
     });
     autoUpdater.on('download-progress', function (info) {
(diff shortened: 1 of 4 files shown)


Jaifroid:
At least fixes update progress. Difficult to test whether the order of notification of update types is correct until this code is already published, as we need to update to latest GitHub version, and THEN notify the user that a PWA update is needed. So, merging for now and will re-open if not fully fixed.

Issue #617, opened 2024-06-16
Uncaught Exception  EARDDRINUSE  address already in use  3000


BridgetteMurray:
Hi All, I've got this error and looking for some help please, thanks

![image](https://github.com/kiwix/kiwix-js-pwa/assets/65484753/f435761f-c643-434f-8f8d-aae0033bf0dc)

Jaifroid:
Hi @BridgetteMurray, it looks like you have some other app that is using port 3000. You can change the port number of the Electron app in configuration, if you can get to the setting. To get to the setting you may need to shut down any other application using port 3000 first (or it could be a duplicate copy of the same app that has got stuck and is hogging the port). A reboot should work, and then it ought to be possible to open the app and change the port number. I would use a number higher than 3010 (it doesn't really matter what, but 3001 and 3002 are reserved for WikiMed and Wikivoyage apps).

Jaifroid:
PS The setting is under "Troubleshooting and Development" in Configuration.

Jaifroid:
#619 Automatically fixes this issue on startup (if port is in use, it increments the port by 10 until a free port is found, and stores that value). However, this version won't be available as an update until the next release, so the workarounds suggested above would need to be used until there is a new release later this month.

BridgetteMurray:
Thanks all, I've identified what is causing the problem, I had an idea but your comments pointed me directly to it, Grafana sits on port 3000, the last couple years I've been doing a lot of assignments with Harvard uni and I'm pretty sure it got loaded to do something with Harvard but I cant remember what. I'm loading Kiwix via Microsoft store which is where the error that I screen shot above comes from, I've used my task manager to end the Grafana service several times but it just pops backup as a different PID still on port 3000, I really don't want to uninstall it just in case something I've worked on that is relying on it. Is there a way I can open Kiwix JS from the Installer application on my HDD, which takes me to the Microsoft store, can I use a command to open on a different Port ?, or do I have to wait for the next version?, thanks

Jaifroid:
@BridgetteMurray Since we now have a nightly version of the app which has the new code in it, we can use that to update the port. According to my experiments, installing and running this will also update the port for your installed Store version, meaning you can then uninstall the nightly after it's done that job. You don't need to uninstall the Store version while doing this. Here's the procedure:

[...]

PR #619, opened 2024-06-16
Add error checking to port assignment

### LINKED ISSUE #618: [Electron app] If port is already in use, automatically scan for a new one on app startup
As reported in #617, the app will crash if the default port (or a new one set by a user) is already in use. We need to catch the exception, increment the port by 10, store the new value, and retry.

Jaifroid:
Fixes #618.


Jaifroid:
Also addresses #617.

Jaifroid:
Tested and working.

PR #635, opened 2024-08-09
Docker size reduction

AllMightLegend:


Jaifroid:
@AllMightLegend Could you kindly explain why this PR is needed and how it reduces the size of the ghcr container?

In general it's a good idea to create an issue first explaining a problem, and then link your PR to that issue. Thanks!

AllMightLegend:
I'm very new to open source contribution.So thanks for the advice.Actually if image is created through alpine the size will be less and less lines of code are always better

docker/dockerfile.pwa
```diff
@@ -1,13 +1,6 @@
...
 
-COPY ./dist/manifest.json /usr/share/nginx/html
-COPY ./dist/replayWorker.js /usr/share/nginx/html
-COPY ./dist/service-worker.js /usr/share/nginx/html
-COPY ./dist/index.html /usr/share/nginx/html
-COPY ./dist/CHANGELOG.md /usr/share/nginx/html
-COPY ./dist/LICENSE /usr/share/nginx/html
-COPY ./dist/www /usr/share/nginx/html/www
-COPY ./dist/archives/README.md /usr/share/nginx/html/archives/README.md
+COPY ./dist .
```
Jaifroid, line 5: Copying everything in dist isn't appropriate, as we may have files for different platforms there that are not relevant to the PWA in particular. It's better to be explicit about what you're copying.

Jaifroid (changes requested):
Please see comments. Thanks!

.history/docker/dockerfile_20240810002439.pwa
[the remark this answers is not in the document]
AllMightLegend: Okay thanks

docker/dockerfile.pwa
AllMightLegend: Oh okay

Issue #572, opened 2024-02-29
In (at least) zimit2 ZIMs, when the content specifies opening a link in a new browser window/tab, the main app window blanks out


Jaifroid:
The content does load in the new tab or window (in browser), and is fully navigable, but there are no app search or navigation bars in that instance. That is as expected, but what should not happen is that the main app UI is blanked out and all navigation controls disappear.

What appears to be happening is that the app is unaware that a new tab has been opened. The Service Worker has no idea of the DOM, so can't tell which client is requesting resources. Because the app believes it is loading a local resource, it tears down the existing article in the iframe ready to receive new content, which instead goes to the new window/tab.

This does not happen if the new tab/window is one opened by the app itself as a result of a user gesture. So, to fix this, we need to handle the fact that the content will be sent by the SW to a different client. We already know how to do this in the message channel event, but it needs to be coordinated.

Jaifroid:
This does **not** happen upstream in KJS. The article does load in a new window/tab, but the main app remains perfectly usable. So it's very much a bug/regression specific to this Repo's earlier implementation.

Jaifroid:
Issue is confirmed to affect only zimit2, not zimit1 archives.

PR #636, opened 2024-08-11
Fix article display race condition

### LINKED ISSUE #634: Race condition causes failure to unhide article contents opened in new container if there is browser-induced delay
I have noticed this on Edge when my PC is offline. An extension or some other code in Edge delays opening of clicked links while it attempts to access the Internet. It blocks the loading of content until this times out. The delay is slight, but it appears to cause a race condition that prevents the unhiding of article contents in new tabs or windows.

I guess that what is happening is a race condition between the code that prevents flash of incorrect style and the code that unhides the hidden content when it is fully loaded. These get out of sync, and the content remains hidden, resulting in a blank window. This does not affect content loaded in the iframe.

Jaifroid:
Fixes #634.

Pretty tricky to catch all cases. Sometimes articleWindow.onload fails to fire, so we have to have a loop to check and launch articleLoadedSW anyway. Then, sometimes , a document is loaded, but not fully at the time the unhide function runs, so we have to have another loop to check, wait, and rerun until we get a positive result. But we also have to avoid creating an infinite loop...

Why are browser events so quirky? 🤬

### DIFF
--- a/www/article.html
+++ b/www/article.html
@@ -5,6 +5,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: blob: file: about: chrome-extension: ms-appx-web: 'unsafe-inline' 'unsafe-eval';">
         <meta name="description" content="Placeholder for injecting an article into the iframe or window">
+        <link rel="icon" href="data:,">
     </head>
     <body></body>
 </html>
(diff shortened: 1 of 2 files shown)


Jaifroid:
I've covered all the edge cases I can detect, with the exception of some abstruse uses of Restricted mode, where we don't officially support all but the most basic usage of opening a separate window or tab. Tested on Chrome, Edge, Firefox (desktop), Edge and Firefox (Android), and iOS Safari 17 on iPad Pro. Note that iPad and Firefox Android do not support opening a clicked link in a new tab or window.

Aside from fixing a bunch of problems where the window on which the user had clicked wasn't correctly detected, and a few timeout issues causing the race condition, this code initiates lazy loading of images more reliably than before, so they will appear to load and be displayed faster.

Jaifroid:
May have some effect on #572, needs testing.

PR #638, opened 2024-08-13
Fix failure to display home or random pages in dynamic ZIMs

### LINKED ISSUE #637: Major issue with Gutenberg ZIMs in PWA (possibly other ZIMs with proprietary UIs)
It's bizarre, but I have an idea what is going on. In the PWA you can navigate well around a Gutenberg ZIM, until you open the HTML version of a book. After that, attempting to go to the Home page (e.g. by clicking on the Home button in the app's own UI), searching for a new page, or even attempting to load a random page, leads to some kind of internal conflict where the app receives the HTML for the book that had been opened instead of the HTML for the requested page. However, it also cannot display that HTML, and instead shows a completely blank page. Inspecting it shows that there is a single URL address in the body of the HTML, not even enclosed in HTML tags. But not the contents of that URL. This is really bizarre.

Restarting the app gets it out of this state.

Here's what I think may be going on: the app has some code that caches the last-visited page for quick startup of the app when resuming the reading of a long article or book. However, much of the navigation in a Gutenberg ZIM is JavaScript-driven via a proprietary (as opposed to client-driven) UI. This appears to lead to the app not registering a navigation when clicking on elements in the UI, which in turn leads both the caching and the browser history management (which was necessary in the days of JQuery mode, but should not be needed now) to enter into a conflicting state. I need to debug to understand it better.

Jaifroid:
Fixes #637. Turns out the wrong contentWindow was being selected. Needs testing in case this regresses #636.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -4953,7 +4953,8 @@ function readArticle (dirEntry) {
     appstate.expectedArticleURLToBeDisplayed = dirEntry.namespace + '/' + dirEntry.url;
     params.pagesLoaded++;
     // We must remove focus from UI elements in order to deselect whichever one was clicked (in both Restricted and SW modes),
-    articleContainer = articleContainer || articleWindow;
+    // articleContainer = articleContainer || articleWindow;
+    articleContainer = (articleWindow.self !== articleWindow.top) ? iframe : articleWindow;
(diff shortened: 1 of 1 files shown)


Jaifroid:
No regression.

PR #641, opened 2024-08-16
Update code signing process

### LINKED ISSUE #628: Update Windows code signing certificate
It's not a bug obviously but it's time sensitive as the current certificate expires in 3 weeks. It's not just a certificate replacement as it's not possible anymore to get a Code Signing certificate with a private key bundled. Private keys now have to be stored in an HSM*.
Our new certificate is thus now using a Cloud Signing solution (we dont have the private key).

All the details are available at https://github.com/kiwix/overview/wiki/Cloud-Code-Signing-(Windows)

⚠️ **IMPORTANT** ⚠️: We have a quota of cloud signs (240 per year for all kiwix binaries) and we decided that we'd **only sign releases** . Signing of nightlies must thus be removed.

Looking at the download folder, it looks like you have 9 files signed per release:

- `Kiwix JS Electron.exe` (inside `kiwix-js-electron_3.3.8_x64.nsis.zip`)
- `Kiwix JS Electron.exe` (inside `kiwix-js-electron_3.3.8.zip`)
- `kiwix-js-electron_x86-64_3.3.8.appx`
- `kiwix-js-windows_3.3.8.appx`
- `kiwix-js-windows_3.3.8.0.appxbundle`
- `kiwix-js-electron_win_setup_3.3.8.exe`
- `kiwix-js-electron_win_portable_3.3.8.exe`
- `kiwix-js-electron_win7_setup_3.3.8.exe`
- `kiwix-js-electron_win_web_setup_3.3.8.exe`

Is this correct? I see that you also publish a lot of variations for Custom Apps. I also see that the release frequency is higher than for other kiwix softwares. @Jaifroid could you provide an estimation of number of signatures needed, per year, for all the ones you are in charge of? We'll update our signing subscription based on your number.

A quick glimpse at your Github Actions workflow makes me think you are not releasing from there. Can you confirm? Can you explain this choice and whether this is subject to change or not. If not, I think we'll provide you with a signing account.

Please get in touch here or on Slack to discuss next steps

Jaifroid:
Fixes #628.

- [x] Fix UWP (legacy app) code signing
- [x] Patch sign process for Electron Builder locally
- [x] Modify workflow for cloud building on GitHub to download and install eSigner dependencies
- [x] Adapt the build script so it is aware when it is running in a GitHub VM
- [x] Build a draft release with artefacts
- [x] Test artefacts


Jaifroid:
It may be worth noting that to sign an appx, it is obligatory that the subject of the certificate should match subject of the appxmanifest contained in the file. In our case, the new certificate has the extra value `S=Canton of Vaud` in its subject, so this was causing failures until fixed. Those failures aren't obvious until you dig into an obscure Windows logfile...

Reference: https://learn.microsoft.com/en-gb/windows/win32/appxpkg/how-to-sign-a-package-using-signtool?redirectedfrom=MSDN

Jaifroid:
Comments on different signtools available on GitHub actions:

* `C:/Program Files (x86)/Windows Kits/10/bin/10.0.22621.0/x64/signtool.exe` - unable to sign any files due to it looking for a private key
* `C:/Program Files (x86)/Windows Kits/10/bin/10.0.22621.0/x86/signtool.exe` - able to sign exe files, but not appx

On my local machine, `C:\Program Files (x86)\Windows Kits\10\bin\x64\signtool.exe` (dated 2015!) can sign all exe and all appx, but it is not available on GH Actions.

Jaifroid:
Tried `choco install -y windows-sdk-10.0`, but this just downloads the SDK 10.0.26624, which evinces the same basic issue: signs `exe` but no `appx`. How come it works fine on my local machine? 🤬

Jaifroid:
No installation SDK needed. What is needed is the latest pre-installed SDK, use the x86 version (bizarrely not the x64 version), ensure for the appx signing that the subject in the appxmanifest exactly matches (including case and spaces) the subject defined in package.json.

Jaifroid:
Note that `.exe` signed with this process will still trigger SmartScreen warnings until the corresponding download has established enough reputation (see issue for more info).

PR #643, opened 2024-08-19
Disable signing for nightlies

### LINKED ISSUE #642: Fix nightly build of Windows packages
#641 prevented code signing on nightly builds, but electron-builder still tries, unsuccessfully, to sign the binaries. See [failed run here](https://github.com/kiwix/kiwix-js-pwa/actions/runs/10446749425/job/28924606833#step:9:81).

The script correctly runs electron-builder with `--publish never`, which should disable signing attempts, but clearly doesn't. Need to find correct command.

Jaifroid:
Fixes #642.


Jaifroid:
It was so hard to get electron-builder to sign correctly in the first place, you'd have thought it would be easy to disable signing. Nothing as simple as a switch, oh no. Here's what worked eventually:

* Created a custom configuration file (electronBuilder.cjs) that:
  + Loads the base configuration from package.json
  + Removes all signing-related properties (certificateSha1, signingHashAlgorithms, rfc3161TimeStampServer, and sign) from the win object
  + Unsets top-level signing configs (cscLink and cscKeyPassword)

This is loaded via two -unsigned package.json scripts that are called conditionally.

PR #657, opened 2024-10-06
Remove title description from popovers

Jaifroid:
This affects at least Wikivoyage ZIMs produced from the mobile HTML endpoint. PR simply applies id-based exclusion filter when constructing the popover content.


Jaifroid:
Tested and working on Firefox, Edge, and Edge 18 via BrowserStack. Also tested that it compiles and runs on IE11.

PR #666, opened 2024-10-31
Use OPFS persistent storage

### LINKED ISSUE #665: Fix Firefox quota issue by requesting persistent Storage type for Origin Private File System
According to a recent update to https://bugzilla.mozilla.org/show_bug.cgi?id=1892994, we can fix the Firefox limited Storage quota by requesting persistent Storage with `navigator.storage.persist()`. I have tried this in dev console on the current PWA, and it does indeed provide unlimited storage, which is great news, and takes away an obstacle to using the Firefox browser on Android to install the PWA.

It needs to be tested on the Android version of course, but this looks very hopeful, and would be a very easy PR I can probably do in the next few days. It should have benefits on Chromium as well, as it may make browser evictions less likely. I have never seen a browser eviction on Chrome for Android, so maybe persistent storage is used by default in that context.

Fixing this issue would make it extremely compelling to address #646 and make use of the Origin Private File System the default on Android, mirroring the upcoming changes in the Kiwix Android app (the mechanism for storing rapid-access files would be nearly identical between the standard Play Store version of the Android app and the PWA).

Jaifroid:
Fixes #665.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1501,16 +1501,21 @@ document.getElementById('archiveFiles').addEventListener('click', function (e) {
     }
 });
 document.getElementById('useOPFSCheck').addEventListener('change', function (e) {
-    var checkAppType = function () {
-        return Promise.resolve(true);
+    var checkStorageType = function () {
+        // Requesting persistent storage (but we'll proceed anyway if the OPFS is available, even if persistence is not granted)
+        if (e.target.checked) {
+            return cache.requestPersistentStorage();
+        } else {
+            return Promise.resolve(true);
+        }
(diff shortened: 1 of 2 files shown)


Jaifroid:
Working well on Firefox for Android on my own device. In Chrome/Chromium, persistence appears to have no effect and is auto-denied, but we continue anyway, and a large quota is still available. It seems that the Chromium implementation of OPFS does not use persistence, but it is crucial for Firefox use at least on Android.

Jaifroid:
I should probably change the log level of persistence messages in console.

Issue #558, opened 2024-02-02
Use `skipWaiting()` in the Electron app to ensure immediate update after upgrade install


Jaifroid:
The Electron app's upgrade process may be a bit confusing for users. With the exe and AppImage versions, it will download updated code in the background while the user is using the app (if they allowed auto-update), and then offer the update, asking users to restart. However, due to the change in architecture whereby the app runs as a (fully local) PWA under the hood, it isn't fully updated after the restart. It will need a FURTHER restart for the PWA to update itself from the new code.

There is a way to upgrade a PWA immediately, using `skipWaiting()`. However, that is not a good experience for standard PWA users because it makes upgrades more intrusive (they can interrupt the browsing session). So we need to be able to detect the situation, which is a bit tricky, but probably achievable.

Jaifroid:
I've investigated this a bit, but calling `skipWaiting()` on its own once the Cache Promises have completed doesn't do the job. It's possible that combining this with an auto-refresh (reload) might work, but it would be important to ensure that it is not obvious or disruptive to the user.

Issue #523, opened 2024-01-01
PWA sometimes fails to cache all its assets, and then does not fetch them when needed


Jaifroid:
If the PWA fails to cache all its assets on first load, it should later fetch (and cache) any resources needed if it is online. But it doesn't always seem to do this. For example, with the `darkreader.min.js` plugin, I have seen a few cases of it failing to load as part of the PWA, and then refusing to load subsequently. It just silently fails and the user is left wondering why dark mode is not responding.

What **should** happen is that resources are fetched from the server whenever needed if they are not cached, as this is part of the PWA specification (fallback to online version). Needs debugging, but it is difficult to reproduce.

PR #668, opened 2024-11-03
Better coding of cache invalidation

Jaifroid:
Experimenting with cache invalidation for reload.


Jaifroid:
Fixes #558 and #523.

PR #674, opened 2024-11-24
Fix vertical positioning of popovers

### LINKED ISSUE #672: Vertical position of popovers no longer adjusting with page zoom factor in PWA!
This is bizarre, because I worked a lot on that, and haven't touched the code, but it now appears that altering the page zoom factor in-app is not adjusting the vertical positioning of popovers, leading them to be positioned over (and hide) the link the user may wish to click.

Grrr...

Jaifroid:
Fixes #672.

### DIFF
--- a/www/-/s/style-dark.css
+++ b/www/-/s/style-dark.css
@@ -1,4 +1,4 @@
-﻿
+
 /* Wikipedia Dark Theme v1.0.23 (1/17/2017)
  * https://github.com/StylishThemes/Wikipedia-Dark
  * http://userstyles.org/styles/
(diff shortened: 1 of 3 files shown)


Jaifroid:
While now working well, seems to have nerfed some aspect of dark mode....

Jaifroid:
Positioning is broken when tabbing into an element. Seems we don't get proper dimensions from the offset values. Additionally, when hovering over a link inside a positioned div (e.g. navbox) the positioning goes haywire.

Jaifroid:
Used a better way to detect the change with getBoundingClientRect(), but needs testing on Safari, UWP and Android.

Jaifroid:
Tested on all common browsers, and also on Edge Legacy. Also Safari and iOS.

As a separate issue, when a page is zoomed excessively, the popovers are rendered too large to fit on a phone screen without horizontal scrolling. But this is minor. It could be dealt with by scaling the size.

PR #683, opened 2025-01-06
Fix search with zoom

### LINKED ISSUE #682: In-app search for contents of page is affected by the `zoom` value
The amount to scroll when searching for words in the currently loaded article (using the custom search function) needs to be multiplied by the zoom factor, as the new `zoom` code affects the scroll to next result function.

Jaifroid:
Fixes #682.


Jaifroid:
Tested and working on:

- [ ] Edge latest
- [ ] IE11
- [ ] Firefox latest
- [ ] Chrome 110
- [ ] Firefox 110
- [ ] Edge Legacy

Issue #708, opened 2025-04-03
Remove ToC when one or less elements


D3V-D:
ToC is irrelevant when only one item (or, sometimes 0-in which case it's an annoyance). Should be as simple as checking how many items we have in the ToC and then choosing whether to show it or not.

Jaifroid:
Thanks for noticing that. Yes, should be simple not to display it if >= 1 element.

PR #710, opened 2025-04-07
Feat:708 Remove ToC when one or less elements

Jainex17:
fixes: #708 

before:
image

After:
wsScreenshot 2025-04-07 184426

## Action:
- In the code for getting the Table of Contents (TOC), we are calling a function twice:
- If the TOC is found on the first call, it will display normally.
- If the TOC length is 0 on the first call, we remove the TOC button and adjust the positions of the other buttons.
- If the TOC is found on the second call, the TOC button will reappear.
- If the TOC is not found in both calls, the TOC button remains removed, and the other buttons stay adjusted.

Note: There is another way to solve this by adjusting the entire navigation based on the TOC value, but it would require too many code changes in the repo. So, I went with this solution instead. If there's anything wrong, please let me know. Thanks


Jainex17:
@Jaifroid Any update?😶‍🌫️

Jaifroid:
My apologies, I'm afraid I'm not on the lookout for volunteer contributions to this Repo, as we prefer people to contribute upstream at https://github.com/kiwix/kiwix-js.

I don't think it's a good idea to remove the button. Instead, the dropup indicator should be removed and the button greyed out so it is not clickable.

Jainex17:
The new PWA has the same implementation it removes the TOC if there is nothing in the dropup.

> My apologies, I'm afraid I'm not on the lookout for volunteer contributions to this Repo, as we prefer people to contribute upstream at https://github.com/kiwix/kiwix-js.

@Jaifroid About this you can close this issue if it's not needed. Thanks

Jaifroid:
@Jainex17 Well, the issue is fine, just I think the button shouldn't be removed, it should just be greyed out and the dropup indicator removed. Are you interested in adapting your PR along these lines?

Jainex17:
Sorry was busy with exams. The background is transparent, so if I grey out the TOC, instead of hiding toc it actually highlights the button. I’ve removed the dropdown indicator and users won’t be able to click on the TOC if there are no items in the dropdown. 

image

Jaifroid (changes requested):
Although the code to disable the button is working, unfortunately it remains disabled even on a page with a valid ToC. See screenshot, original app on left, your code on right.

Jaifroid:
Here's the screenshot:

image

Jaifroid:
Also, please update the branch with latest code from main (use Update branch button below).

Jainex17:
@Jaifroid It's working fine for me. Could you please provide the document you are using?

Screenshot 2025-06-16 200037

Jainex17:
@Jaifroid Just following up could you share the document you're using so I can help troubleshoot?

Jaifroid:
Instructions to reproduce:

1. Download https://download.kiwix.org/zim/wikipedia/wikipedia_en_medicine_nopic_2025-03.zim
2. Open file with the app checked out for this branch
3. Ensure you are on the Home page of the app - because this page has no headings, the icon is correctly greyed out
4. Navigate to any other page with headings - button is still greyed out.

I just checked that, and the issue is still there.

Issue #711, opened 2025-04-10
Add explicit support for new ZIMs without title listing v0 (minor version 3)


Jaifroid:
See https://github.com/openzim/libzim/issues/800. While the PWA supports this ZIM type via an appropriate assumption, it spits out an error in console to say it does not recognize minor version 3: `Unknown ZIM minor version: 3! Assuming content namespace is C.`

We should add explicit support for this version of ZIM, once the outcome of the linked issue is known. The support should be backported upstream to the browser extension.

PR #713, opened 2025-04-18
Accept ZIM minor version 3

Jaifroid:
This PR fixes failure to detect the new ZIM minor version 3 that has no v0 title index (see https://github.com/openzim/libzim/issues/800 for full details).

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -468,13 +468,13 @@ ZIMArchive.prototype.findDirEntriesWithPrefix = function (search, callback, noIn
 ZIMArchive.prototype.getContentNamespace = function () {
     if (this.isReady()) {
         var ver = this.file.minorVersion;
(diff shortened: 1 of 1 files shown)


Jaifroid:
See also https://openzim.org/wiki/ZIM_file_format#Major_.26_Minor_versions, where minorVersion 3 is characterized as removing title index "listing/titleOrdered/v0" (old zimcheck will complain). Since we find listings from highest to lowest, there is no impact on our low-level reading of the ZIM format.

Jaifroid:
I have manually tested with a test ZIM that has minorVersion 3:

image

No warning is shown. We do not have automated tests for such a ZIM, as this would require writing a whole new test suite for a new ZIM, which is overkill for testing such a low-impact, easily fixable issue.

Jaifroid:
Tested and working on ZIMs with minorVersions 0, 1, 2, 3.

Jaifroid:
Fixes #711.

PR #717, opened 2025-05-21
Fix actionparse wikimedia custom styling

Jaifroid:
This fixes specific display issues in new actionparse ZIMs.


Jaifroid:
Appears to be working.

PR #725, opened 2025-06-10
Add snippets to fulltext search

### LINKED ISSUE #724: Add snippets to fulltext search
Now that we have a libzim with snippet support, we need to backport the implementation from upstream (Kiwix JS) to this app.

Jaifroid:
Closes #724.


Jaifroid:
The implementation is currently limiting title search to 15 scans, which often fails to find the precise title searched for. On main, this isn't broken, so something about the rate-limiting code for costly fulltext search is interfering with the highly optimized code for binary search. This definitely needs fixing before merge. We may need to specify a max for ft search that is separate from the max search window for binary search. The latter should keep going until it finds the required number of titles.

PR #784, opened 2025-09-18
Add persistent permissions handler for FileSystem

### LINKED ISSUE #783: Add support for new FileSystem permission check handler in Electron
This is available in Electron nightly (40+), and should be published and backported to 35 at least.

As this was the main obstacle to updating modern Windows apps to latest Electron, we should add the appropriate handler in main.js.

See: https://github.com/electron/electron/issues/41957

Jaifroid:
Fixes #783.


Jaifroid:
This can't be merged until Electron 40 is out. We don't have to peg to Electron 40, as the fix is backported to some earlier Electron versions, but those won't be released until Electron 40, if I understood correctly.

PR #790, opened 2025-10-15
Add keyboard shortcut for search (in ZIM) and fix keyboard event listeners

### LINKED ISSUE #788: Cannot acces search bar using keyboard shortcut
I am on Firefox 143 and maxOs 15.
When using the PWA, the search bar states "Search [.*] or space key..."
Reading this, I assume that the space key can be used as a keyboard shortcut to reach the search bar.

However, this does not work the space bar is scrolling down the page in increments instead.

Jaifroid:
Fixes #788 using '/' as shortcut to focus the search field (prefix). Also adds support for macOS Command key modifier for Cmd+P, and Cmd+F for printing and find-in-article.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -4878,13 +4878,15 @@ function listenForSearchAndPrintKeys (controlWindow) {
     controlWindow.addEventListener('keydown', function (e) {
         // Ctrl-P to patch printing support, so iframe gets printed
         if ((e.ctrlKey || e.metaKey) && e.key.toLowerCase() === 'p') {
-            e.stopPropagation();
             e.preventDefault();
+            e.stopPropagation();
             printIntercept();
         }
-        // Focus the prefix if user types a / (but not in an input field)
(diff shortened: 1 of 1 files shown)


Jaifroid:
I've implemented the following:

**Universal shortcuts (should work in all contexts):**
* **/** (forward slash) - as suggested by user in issue they created
* **Ctrl+K** (or **Cmd+K** on Mac) - following modern app conventions
* **Home** key - semantic "start here" shortcut

**Installed PWA or Electron apps only:**
* Ctrl+L / Cmd+L - traditional browser address bar shortcut
* Alt+D - alternative traditional address bar shortcut

These last two won't work when running the app in a regular browser tab (they'll focus the browser's address bar instead), but they work great in a PWA installed (in Chrome/Edge or other Chromium browser) or in the Electron app, where there's no browser chrome.

Jaifroid:
Added support for some edge cases of select elements or contenteditable other than an input field (such as in freecodecamp). In these cases, Home and / do not focus the search field.

PR #800, opened 2025-11-18
docs: add fallback PWA installation guide (fix #626)

### LINKED ISSUE #626: Add documentation explaining how to install the PWA if `onBeforeInstall()` event is not supported by the browser
As an outcome of #625, I think I should add some documentation explaining how to install / add to home screen the PWA in browsers that don't support the install event. Although all Chromium browsers should support it, in some cases (e.g. Vivaldi) the function appears to be supressed for security reasons, and in others, like Firefox on Android or Safari on iOS, it is not supported even though the app can in fact be added to Home and can work as an independent app without any browser chrome.

Bharatt16:
This PR adds missing documentation explaining how to install/add-to-home-screen the Kiwix JS PWA in browsers that do not support the beforeinstallprompt install event, or where the install prompt is intentionally suppressed (e.g., Safari on iOS, Firefox for Android, Vivaldi).
This documentation was requested as the outcome of issue #626.

🔍 Before (Problem / Missing Content)

Previously, the README only explained how to install the PWA when the install event is supported (mainly Chromium browsers).
However:

-> Safari on iOS never fires the install event

-> Firefox for Android does not trigger the event

-> Vivaldi (Chromium-based) suppresses the event for security reasons

-> Some browsers partially support PWA install UI but not the event

-> Users in these browsers did not receive clear guidance on how to install the PWA manually

As a result, users could still install the PWA, but no documentation explained how to do it when the install prompt does not appear.

🛠️ What This PR Adds (Solution)

✔ A new file:
readme/PWA_FALLBACK_INSTALL.md

This file provides:

-> A clear explanation of why the install event may not fire

-> Platform-specific installation instructions for:

        -> Android (Chrome/Edge/Brave/Samsung Internet/Firefox/Vivaldi)

        -> iOS/iPadOS (Safari)

        -> Desktop (Windows/macOS/Linux in Chrome/Edge/Brave)

        -> Firefox (where PWA install support varies)

-> Steps for manually adding the PWA to Home Screen or installing via browser UI

-> Ensures users know how to install the PWA even without automatic prompts

✔ A link in README.md under the PWA section, pointing to this new documentation so users can easily find it.

📈 After (What It Is Now / Impact)

-> The PWA documentation is now complete and covers the full range of browsers.

-> Users who do not see an install prompt now have a clear, official guide.

-> Browsers that suppress or lack PWA install events are now properly documented.

-> Prevents confusion for new users and improves accessibility across all platforms.

Resolves the request in Issue #626.

🔗 Related Issue

This PR fixes: #626

### DIFF
--- a/screenshots/Install-PWA.md
+++ b/screenshots/Install-PWA.md
@@ -12,7 +12,7 @@ See also: [&ensp;[File handling](Demo-FileHandling.md)&ensp;]  [&ensp;[Install P
 Some browsers do **not** display the PWA install prompt (Safari, Firefox, Vivaldi, etc.).  
 If the “Install App” button does not appear, follow:
 
-👉 **[PWA Fallback Installation Guide](../PWA_FALLBACK_INSTALL.md)**
+**[PWA Fallback Installation Guide](../PWA_FALLBACK_INSTALL.md)**
 
 This explains how to install / Add to Home Screen on iOS, Firefox, Vivaldi, desktop Linux, etc.
 


Jaifroid (commented):
Thanks, that's much better. Could you also link to this guide from screenshots/Install-PWA.md? That's what's linked from the main README.md (in root directory), so it would be good to have the alternative installation method linked from the paragraph about how to install underneath the video.

Bharatt16:
Done! I’ve linked the fallback installation guide in screenshots/Install-PWA.md as requested. Let me know if you'd like any adjustments.

PR #843, opened 2026-04-24
Update release/nightly uploads for new destination

rgaudin:
As part of a server/infra change, the upload of nightly and release builds is done to a new destination.

The receiving service now uses per-project (well per-repo) users with independent SSH Keys. Each users now only has access to the folders in which it is expected to upload files.

- Host remains the same: `master.download.kiwix.org`
- Username changes from `ci` to `kiwix-js-pwa`
- Port changes from `30022` to `30322`
- Paths remains the same

⚠️ DO NOT MERGE ATM

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -72,7 +72,7 @@ env:
   ESIGNER_TOTP_SECRET: ${{ secrets.ESIGNER_USER_TOTP_SECRET }}
   MASTER_KEY_FILE: "C:\\Users\\runneradmin\\eSignerCKA\\master.key"
   INSTALL_DIR: C:\Users\runneradmin\eSignerCKA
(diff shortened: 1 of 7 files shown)


Jaifroid:
@rgaudin Looks like this is failing for the Windows builds despite the fixup PR yesterday. See https://github.com/kiwix/kiwix-js-pwa/actions/runs/25537153851/job/74955137988 .

Can you see an obvious cause? Otherwise, I'll take a look when I can.

By the way, I'm happy to change the file location to kiwix-js-pwa, but it needs co-ordination with Popolechien to be sure the links in kiwix.org are correct.

rgaudin:
Sorry took me a couple of trials before getting it right. Workflow eventually completed correctly. Thank you for reporting!

Jaifroid:
Thank you, the nightly just completed with no error, so all good! 🥳

Issue #874, opened 2026-07-15
Update Windows code signing certificate / secrets (expired 11th July)


Jaifroid:
The Kiwix code signing certificate expired on 11th July. Have requested access to the new version for 2026-27.

When I have access, I will need to update at least the `ESIGNER_USER_TOTP_SECRET`.

Jaifroid:
See https://github.com/kiwix/operations/issues/445 for further info.

PR #878, opened 2026-07-19
Add web torrent for Electron packages

### LINKED ISSUE #877: Add Node-based WebTorrent support to the Electron apps
In line with https://github.com/kiwix/libkiwix/issues/1193 and https://github.com/orgs/kiwix/projects/27/views/1, it would be viable to add the WebTorrent dependency (the Node JS version) to the Electron and NWJS apps.

Much of the infrastructure already exists. I don't consider this makes sense for the PWA, as that would only be able to see WebRTC peers, and I believe Kiwix doesn't announce or serve files via WebRTC. But there is no such limitation for Node JS, and [WebTorrent](https://github.com/webtorrent/webtorrent) provides a full-featured library for Node that we could leverage relatively easily.

Jaifroid:
Fixes #877. This uses WebTorrent under NodeJS, so that it can interact with the full BitTorrent protocol, rather than just the WebRTC hosts that are supported under plain JS: This means that only Electron and NWJS packages will leverage BitTorrent downloading.

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -881,7 +881,7 @@ jobs:
                     if (!$INPUT_SIGN) {
                         npx electron-builder build --config ../scripts/electronBuilder.cjs --win nsis-web --ia32 --x64 --arm64 --publish always --projectDir dist
                     } else {
-                        npx electron-builder build --win nsis-web --ia32 --x64 --arm64 --publish always --projectDir dist
+                        npx electron-builder build --config ../scripts/electronBuilder-signed.cjs --win nsis-web --ia32 --x64 --arm64 --publish always --projectDir dist
                     }
                 }
                 'appx' {
@@ -899,7 +899,7 @@ jobs:
                     } else {
                         npx electron-builder build --win $PHASE1_TARGETS --publish always --projectDir dist
                         node scripts/delete-latest-yml.js
-                        npx electron-builder build --win nsis-web --ia32 --x64 --arm64 --publish always --projectDir dist
+                        npx electron-builder build --config ../scripts/electronBuilder-signed.cjs --win nsis-web --ia32 --x64 --arm64 --publish always --projectDir dist
                     }
                 }
             }
@@ -910,7 +910,7 @@ jobs:
                     if (!$INPUT_SIGN) {
                         npx electron-builder build --config ../scripts/electronBuilder.cjs --win nsis-web --ia32 --x64 --arm64 --projectDir dist
                     } else {
-                        npx electron-builder build --win nsis-web --ia32 --x64 --arm64 --projectDir dist
+                        npx electron-builder build --config ../scripts/electronBuilder-signed.cjs --win nsis-web --ia32 --x64 --arm64 --projectDir dist
                     }
                 }
                 'appx' {
@@ -928,7 +928,7 @@ jobs:
                     } else {
                         npx electron-builder build --win $PHASE1_TARGETS --projectDir dist
                         node scripts/delete-latest-yml.js
-                        npx electron-builder build --win nsis-web --ia32 --x64 --arm64 --projectDir dist
+                        npx electron-builder build --config ../scripts/electronBuilder-signed.cjs --win nsis-web --ia32 --x64 --arm64 --projectDir dist
                     }
                 }
             }
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -180,11 +180,23 @@ jobs:
           PUBLISH_FLAG="always"
         fi
         echo "Building 64bit packages for ref_name=$REF_NAME..."
+        # Build the x64 targets first, using the runner's native linux-x64 node-datachannel binary.
+        # electron-builder has npmRebuild:false, so it packages whatever single binary is on disk
+        # into every arch it builds — building arm64 in the same pass would ship the x64 binary in
+        # the arm64 AppImage (dlopen would fail on ARM). So after the x64 targets, swap in the
+        # linux-arm64 prebuild (a plain N-API download via prebuild-install, no compilation) and
+        # build the arm64 AppImage separately. The node -e form invokes prebuild-install from the
+        # package dir so it reads node-datachannel's own package.json, hoisting-agnostic.
         if [[ $REF_NAME = "main" ]]; then
-          npx electron-builder --linux AppImage:x64 AppImage:arm64 deb:x64 rpm:x64 --publish $PUBLISH_FLAG --projectDir dist
+          npx electron-builder --linux AppImage:x64 deb:x64 rpm:x64 --publish $PUBLISH_FLAG --projectDir dist
         else
-          npx electron-builder --linux AppImage:x64 AppImage:arm64 deb:x64 --publish $PUBLISH_FLAG --projectDir dist
+          npx electron-builder --linux AppImage:x64 deb:x64 --publish $PUBLISH_FLAG --projectDir dist
         fi
+        echo "Swapping node-datachannel to its linux-arm64 prebuild for the ARM AppImage..."
+        ( cd dist/node_modules/node-datachannel && node -e "process.argv.push('-r','napi','--arch=arm64','--platform=linux'); require('prebuild-install/bin.js')" )
+        file dist/node_modules/node-datachannel/build/Release/node_datachannel.node
+        echo "Building arm64 AppImage (with the linux-arm64 binary swapped in)..."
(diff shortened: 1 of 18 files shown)


Jaifroid:
## Progress summary and development notes

A record of what has been done on this branch so far and why, since several of the fixes address subtle platform issues that are worth remembering.

### Architecture

- BitTorrent support runs in the **Electron main process** via `torrentDownloader.cjs`, which lazily imports WebTorrent (an ESM package) on first use. The renderer communicates with it over IPC exposed in `preload.cjs`, so no Node integration is needed in the renderer.
- Torrent downloads are offered from the download library UI: under Electron, a ZIM's torrent/magnet is downloaded in-app with full BitTorrent peer support (not just the WebRTC peers plain-JS WebTorrent can reach).
- Torrents are saved directly into the user's picked archive folder, so a completed archive shows up in the archive list on refresh, and completed archives continue seeding.

### Main issues encountered and resolved (in commit order)

1. **Initial wiring** (71452a7): WebTorrent backend, IPC bridge, download UI and progress reporting for Electron.
2. **Windows drive-root saves** (70a3b22): a bare drive letter like `W:` is *drive-relative* in Node, so downloads to a drive root went astray; paths are now normalised with a root slash.
3. **Firewall diagnostics** (2826d5d): clearer error message when the OS firewall blocks torrent traffic.
4. **Save path from picked folder + FSA handle hardening** (8b27659): under Electron ≥32, File objects from the File System Access API no longer expose `.path`, so the app has to track *both* real paths (from the native folder dialogue / stored settings) and FSA directory handles (IndexedDB). The torrent save path is now derived from whichever is available.
5. **Packaged build fixes** (a8670ff): skip native module rebuild (`npmRebuild: false`) and ship `torrentDownloader.cjs` in the electron-builder file list — the packaged app previously lacked the torrent backend.
6. **Verification, status routing, folder change** (8fc1767): verify torrent data already on disk before resuming; route status/progress messages by torrent so concurrent downloads don't cross streams; allow changing the download folder.
7. **Seeding vs reading** (1f169d7): a completed archive that is still seeding is reopened **read-only** so that seeding and reading the ZIM don't conflict over file locks on Windows.
8. **Node-fs-first archive loading** (130380b, 102652e, 542cf33, and part of c7f1fad): Chromium File snapshots obtained through FSA handles can be invalidated on Windows by concurrent access to the underlying file (e.g. while seeding), producing garbled reads. The app now prefers Node `fs` loading whenever the current folder is known by a *verifiable real path*. Two regressions from the first attempt were fixed: the stored path could silently substitute a same-named (possibly partial) copy from a different folder — the Node route is now taken only when `params.pickedFolder` is a genuine path string and the file's existence is `stat`-verified; and the archive list no longer collapses to a single entry when loading from the list.
9. **Drag-and-drop flag leak** (c7f1fad): `appstate.filesDropped` was only cleared when the dropped file had a real path; an FSA drop left it set, so every later load from the archive list wrongly collapsed the list and voided the picked folder. The flag is now always cleared on the load following a drop.
10. **CVE-2024-29415 / CodeFactor fix** (b21dc1c): trivy flagged the transitive `ip@2.0.1` (webtorrent → torrent-discovery → bittorrent-tracker). No fixed `ip` release exists, so an npm `overrides` entry pins it to the patched fork **`neoip@2.1.0`**. Version 2.1.0 rather than 3.x is deliberate: bittorrent-tracker default-imports `ip`, and neoip 3.x's ESM build has only named exports, which breaks `import('webtorrent')` at load time. Verified: `npm audit` clean, WebTorrent import chain intact, and the SSRF misclassification (`isPrivate('127.1')`, `isPrivate('0x7f.1')`) fixed. The vulnerable code path (`lib/server/parse-udp.js`) is tracker-*server* code the app never runs, so this was CI hygiene rather than a live exploit.
11. **Cross-architecture native binary fix** (macOS b799523, legacy-macOS 1e6e5ee; Linux 439657f; Windows 0c3164c; incidental 07948c3): WebTorrent's WebRTC transport needs the native `node-datachannel` addon, and because `npmRebuild: false` makes electron-builder copy the single binary from the x64 CI runner into *every* arch package, non-x64 packages shipped an x64 binary that fails to load with an architecture-mismatch error. **macOS and Linux** are built per-arch in separate workflow steps, so each downloads and swaps in the correct `node-datachannel` prebuild between builds (macOS arm64/x64; Linux x64/arm64). The **Windows `nsis-web`** installer must build `ia32/x64/arm64` in a single pass (its Web Setup `.exe` detects the arch at install time and pulls the matching package), so it can't be split; instead a per-arch `afterPack` hook (`scripts/afterPack.cjs`) downloads the correct `win32-<arch>` prebuild, overwrites the packed binary, and **verifies its PE machine type, failing the build on any mismatch** — the CI-side guarantee, since Windows-on-ARM can't be run in this environment. It is wired as a function via `scripts/electronBuilder.cjs` (unsigned) and `scripts/electronBuilder-signed.cjs` (signed, signing config left intact). `ia32` has no `node-datachannel` prebuild and the feature is gated off there (see `preload.cjs` `torrentSupported`), so its unusable binary is simply removed. An incidental electron-builder quirk was fixed alongside (07948c3): in the combined phase-1 `--win NSIS:ia32 portable:ia32 appx` list, the bare `appx` target inherited `ia32` from the other targets and aborted the build with "Unsupported arch ia32 appx", so it is now pinned to `appx:x64`.

### Platform support matrix

In-app BitTorrent is available only where the runtime can host WebTorrent 3.x (Node 20+) **and** a `node-datachannel` prebuild exists for the arch (i.e. not 32-bit `ia32`). Everywhere else the feature auto-hides. Verified per package below (✔ = feature works; ✖ = feature correctly absent):

- [x] **Linux x64** — ✔ available (tested)
- [x] **Linux arm64** — ✔ correct `linux-arm64` binary swapped in (verified by construction)
- [x] **Linux ia32** — ✖ webtorrent stripped from the pinned legacy build (tested)
- [x] **macOS arm64** — ✔ available and working (tested on-device)
- [x] **macOS x64** — ✔ available and working (tested under Rosetta on Apple Silicon)
- [x] **macOS High Sierra / Mojave** — ✖ webtorrent stripped from the legacy build (tested)
- [x] **Windows 11 x64** — ✔ available and working (tested locally and in CI)
- [x] **Windows 11 arm64** — ✔ correct `win32-arm64` binary in package, PE `0xaa64` verified in the CI build log (not run on-device; no Windows-on-ARM hardware available)
- [x] **Windows 11 ia32** — ✖ binary removed, feature gated off (tested)
- [x] **Windows 7 ia32** — ✖ webtorrent stripped from the legacy Win7 build (tested)

> **Signed Windows path:** the signed `nsis-web` build (via `electronBuilder-signed.cjs`) uses the *identical* `afterPack` hook as the verified unsigned build, differing only in that it keeps signing enabled. It has not yet been exercised end-to-end because the SSL.com code-signing certificates are expired and the eSigner step fails before any package is built (new certs pending from SSL.com). It is therefore verified-by-construction until the certs land.

### Remaining to do

- [x] **Resume across restart**: optionally prompt to resume incomplete torrent downloads when the app relaunches.
- [x] Wider testing of packaged builds across all platforms and architectures (see matrix above).
- [x] Update the PR description to reflect the final scope before review/merge (this comment; see item 11 and the support matrix).
- [ ] Signed Windows release run, once replacement SSL.com certificates are available.

Jaifroid:
Current UI:

image

Jaifroid:
**NWJS support for in-app BitTorrent: deferred (with rationale)**

In-app BitTorrent now works in **Electron**. NWJS is intentionally **not** wired up, and the reason is architectural rather than a fixable bug:

WebTorrent's transport stack eagerly loads the native WebRTC addon `node-datachannel` at import time (a static ESM import that no `try/catch` can intercept). In NWJS, Node runs *inside* the Chromium renderer, and that native addon **aborts the renderer** on load (the STATUS_BREAKPOINT crash). Electron is immune only because its torrent client runs in the **main** process — a clean, pure-Node context the renderer never touches.

The **only** viable NWJS path is to reproduce that separation: run `torrentDownloader.cjs` in a real Node context outside the renderer (an NWJS `node-main` script or a spawned child Node process, bridged by IPC). That's a whole second process-bridge architecture, NWJS-only. And even done, the x86/XP builds are permanently excluded — `node-datachannel` ships no ia32 prebuild, and NWJS 0.14.7 (Node 5.11) can't load modern ESM or native modules at all.

Given NWJS already has working **direct download**, and the effort/audience/maintenance trade-off is poor, torrents are deferred there. NWJS continues to receive direct download only.

Jaifroid:
Further progress is blocked, waiting on #874 and https://github.com/kiwix/operations/issues/445.

PR #914, opened 2026-08-08
 Fix #906: decouple slide-away toolbar from #top

### LINKED ISSUE #906: Decouple the slide-away toolbar from #top, and give it an explicit state flag
(Post made by Claude)

## Summary

The slide-away toolbar transforms `#top`, the header element that contains *everything* at the top of
the app. That is fine while every child of the header is supposed to slide, but it stops being true as
soon as one of them is not — as the emulated title bar added in #905 now demonstrates. The header is
also carrying its own visibility state in the string value of its inline `transform`, which is read back
by regex in two places. Both should be fixed together, because the second is what makes the first
awkward to do.

To be picked up once #905 is merged; the line and file references below are to the state of the code
after that PR.

This is tech debt with a known blast radius, not a bug: the app behaves correctly today. Filing it so the
constraints are written down while they are fresh.

## Background: what the title bar work had to do

When Chromium draws the Window Controls Overlay, the app is given the whole window and there is no way to
ask for the real title bar back, so the app draws one itself (`#wcoTitleBar`, a flex line at the top of
`#navbar`). A title bar is window chrome: it should not scroll away, and once the toolbars are hidden it
is the only part of the app the user can grab to move the window.

Because it lives inside `#top`, it cannot be pinned declaratively:

* `position: sticky` needs a scrollport, and nothing here scrolls — the toolbar is moved by a transform
  and the article lives in an iframe. Sticky therefore computes to plain `relative` and never moves.
* `position: fixed` is positioned against the nearest transformed ancestor, and `#top` always carries a
  transform (even `translateY(0)`), so a fixed child travels with the header regardless.

So `hideSlidingUIElements` pushes the bar back down by exactly the distance the header moves up. The two
transforms are on the same curve and start in the same frame, so they sum to zero and the bar is visually
motionless. That is exact, not approximate — but it rests on an invariant that nothing enforces.

## The invariant

1. **The two transitions must stay identical.** `#top` and `#wcoTitleBar` share one declaration in
   `www/css/app.css` (`transition: transform 0.5s ease`). Change the duration or easing for one without
   the other and the cancellation stops being exact.
2. **Both inline styles must be set in the same frame.** Any refactor that defers one of them — a
   `requestAnimationFrame`, a debounce, an early return between the two assignments — breaks it.
3. **Every path that restores the header must also release the bar.** There are two today:
   `showSlidingUIElements` in `www/js/lib/uiUtil.js`, and the `btnTop` click handler in `www/js/app.js`,
   which sets `header.style.transform` directly rather than going through `uiUtil`.

(1) and (2) fail visibly but only *during* a 0.5s animation — the bar lurches to its offset and then
flies back into place — which is easy to introduce and easy to miss in testing. (3) fails persistently:
a third restore path added later would strand the title bar below the header, over the article, and
nothing in the code would catch it.

## Proposed change

**Part 1 — move the transform inward.** Slide `#navbar > .container-fluid` (the search row and the
find-in-article row) instead of `#top`. The title bar is then simply not transformed by anything, so:

* the counter-transform in `hideSlidingUIElements` goes away, and with it all three parts of the
  invariant above;
* the shared `transition` declaration no longer has to be shared;
* `showSlidingUIElements` and the `btnTop` handler have nothing to release.

The `z-index: 1` on `#wcoTitleBar` should stay: the row still needs to pass *behind* the bar rather than
over it, and Bootstrap positions the controls in that row (`.btn`, and the children of `.input-group`,
are all `position: relative`), so without it they paint in tree order over the bar.

The article geometry in `hideSlidingUIElements` — sliding the article up by `headerHeight - titleBarHeight`
and taking the same off its height — stays as it is, since the article still has to start below the bar.

**Part 2 — replace the visibility flag.** The toolbar's hidden/shown state is currently inferred by
testing the header's inline transform string against `/\(0p?x?\)/`, in `www/js/lib/uiUtil.js` (in
`slideAway`) and in `www/js/app.js` (in the `btnTop` handler). This has to be replaced with an explicit
flag — a property on `appstate`, or a class on `documentElement` — before Part 1 can happen, because the
transform being tested is the one that is moving. It is worth having on its own merits: parsing an inline
style string for application state is fragile for reasons that have nothing to do with title bars.

## Scope

`showSlidingUIElements` has eight call sites in `www/js/app.js` plus three uses inside `uiUtil.js`;
`hideSlidingUIElements` has one call site in `app.js` plus two in `uiUtil.js`. All of them need
re-verifying against the new flag, since several restore the toolbar as a side effect of something else
(navigating home, opening configuration, entering and leaving full screen).

Worth checking while in there: the `btnTop` handler restores the header without restoring the article's
transform or height, which `showSlidingUIElements` does. That may be deliberate, but it looks like an
inconsistency.

## Verification

With the Window Controls Overlay drawn (an Edge or Chrome PWA install) and **Show a title bar** enabled
in Configuration:

* scroll down — the search row slides away behind the title bar, which does not move by so much as a
  pixel, at any point in the animation;
* scroll back up — the row returns, with no residual offset on the bar;
* press the **Top** button while the toolbar is hidden — the toolbar returns and the bar stays put;
* drag the window by the title bar in the scrolled state;
* repeat with **Show a title bar** off, and again in a normal browser tab, in Electron and in NW.js,
  where `#wcoTitleBar` measures zero and behaviour must be pixel-identical to today.

## Dependencies

None on the browser side. This is worth doing whether or not Edge restores the user-facing control over
the overlay: if it does, fewer users will turn the title bar on, but the code and its invariant still
ship, and Part 2 stands on its own regardless.

shashwat-28410:
### Summary

  Slide `#navbar .container-fluid` instead of the whole header (`#top`), so `#wcoTitleBar` (the emulated window-controls title bar)
  never moves and needs no counter-transform. Replace the fragile regex-based parsing of the inline transform string with an
  explicit `appstate.toolbarVisible` flag as the single source of visibility state.

  ### Background

  The slide-away toolbar used to transform the entire `#top` header off-screen. `#wcoTitleBar` lives inside `#top` and must remain
  fixed — it's window chrome and the only grab handle to move the window when toolbars are hidden. The old code held it in place
  with a coordinated counter-transform that relied on three unenforced invariants (identical transition curves, both transforms set
  in the same frame, and two restore paths kept in sync). `#wcoTitleBar` and `.container-fluid` are actually siblings, so
  `#wcoTitleBar` never moves if `.container-fluid` slides alone — the counter-transform (and all its fragile invariants) disappears
  entirely.

  ### Changes

  | File | Change |
  |---|---|
  | `www/js/init.js` | Add `appstate.toolbarVisible = true` — single source of truth for toolbar visibility |
  | `www/css/app.css` | Transition selector: `#top`, `#wcoTitleBar` → `#navbar .container-fluid` |
  | `www/js/lib/uiUtil.js` | `hideSlidingUIElements()` / `showSlidingUIElements()` rewritten to transform `.container-fluid` only;
  `slideAway()` reads the flag instead of regex-testing the inline transform |
  resetting transforms |
  
  ### Blast radius
  
  - Visibility state tracked via `appstate.toolbarVisible` instead of parsing the header's inline `transform` string.
  - Restore paths consolidated: `btnTop` now delegates to `showSlidingUIElements()`, so there's a single path to maintain.
  - Title bar pinning relies on the sibling DOM structure (`#wcoTitleBar` beside `.container-fluid` inside `#navbar`), which holds
  regardless of scroll context.
  
  ### Manual verification
  
  Requires a Chromium WCO environment (installed PWA or Electron) to exercise the `#wcoTitleBar`:
  - Scroll down in an article: the search row slides up; `#wcoTitleBar` stays pinned at the top.
  - `btnTop` while the toolbar is hidden restores it.
  - Smooth transition, no flicker (preserved `z-index: 1` on `#wcoTitleBar`).

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -171,10 +171,11 @@ div:not(.card.border-success .card-body):not(.alert-message) {
     overflow-wrap: break-word;
 }
 
-/* #wcoTitleBar is held against the top of the window by a transform that cancels the one sliding #top
-   away (see hideSlidingUIElements), so it has to be animated on exactly the same curve over exactly the
-   same time, or it lurches to its offset while the header eases into place and the bar appears to fly */
-#top, #articleContent, #footer, article, #wcoTitleBar {
+/* #navbar .container-fluid is the slide-away target (see hideSlidingUIElements): it slides out from
+   under the title bar rather than taking the whole header with it, so #top and #wcoTitleBar no longer
+   need a coordinated transition. These three elements animate on the same ease so the layout shifts
+   smoothly together */
+#articleContent, #footer, article, #navbar .container-fluid {
     transition: transform 0.5s ease;
 }
 
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -889,13 +889,10 @@ if (typeof Windows !== 'undefined' &&
 }
 
 document.getElementById('btnTop').addEventListener('click', function () {
-    var header = document.getElementById('top');
     var iframe = document.getElementById('articleContent');
     // If the toolbar is hidden, show it instead of jumping to top
-    if (!/\(0p?x?\)/.test(header.style.transform)) {
-        header.style.transform = 'translateY(0)';
-        // Release the emulated title bar, which hideSlidingUIElements holds down against the header
-        document.getElementById('wcoTitleBar').style.transform = '';
+    if (!appstate.toolbarVisible) {
+        uiUtil.showSlidingUIElements();
     } else {
         if (!params.hideToolbars) iframe.style.transform = 'translateY(-1px)';
         iframe.contentWindow.scrollTo({
--- a/www/js/init.js
+++ b/www/js/init.js
@@ -67,6 +67,11 @@ var params = {};
  */
 var appstate = {};
 
+// Track whether the slide-away toolbars are currently visible. This is the single source of truth for
+// the toolbar visibility state, replacing the previous fragile approach of parsing the header's inline
+// transform string with a regex (see uiUtil.js hideSlidingUIElements/showSlidingUIElements)
+appstate.toolbarVisible = true;
+
 // ******** UPDATE VERSION IN service-worker.js TO MATCH VERSION AND CHECK PWASERVER BELOW!!!!!!! *******
 params['appVersion'] = '3.8.9'; // DEV: Manually update this version when there is a new release: it is compared to the Settings Store "appVersion" in order to show first-time info, and the cookie is updated in app.js
 // ******* UPDATE THIS ^^^^^^ IN service worker AND PWA-SERVER BELOW !! ********************
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -40,6 +40,9 @@ const footer = document.getElementById('footer');
 // browser is drawing the window controls overlay over the app and the user has asked for a title bar of
 // the app's own (see the html.show-titlebar rules in app.css)
 const titleBar = document.getElementById('wcoTitleBar');
+// The inner container that holds the search rows — this is the element that slides away, keeping the
+// title bar and the navbar shell fixed in place (see hideSlidingUIElements / showSlidingUIElements)
+const containerFluid = header.querySelector('.container-fluid');
 let articleContainer = document.getElementById('articleContent');
 
 /**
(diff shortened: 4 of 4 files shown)


Jaifroid:
@shashwat-28410 Thaks for the PR. Could you please put a quick comment in #906 mentioning this PR so that I can assign the issue to you?

shashwat-28410:
Hi — I've been assigned this. The code is ready for review whenever you get a chance. Happy to make any changes you'd like.

Jaifroid:
Thanks for this PR, @shashwat-28410 . However, when I tested it I ran into issues immediately:

image

* When scrolling down, the navbar slides away, but leaves a blank area at the top
* Changing theme from the OS default to light or dark is broken in your PR

It looks like you coded this with an LLM, but never actually set up the Repo and tested the implementation. Please understand that we don't accept vibe-coded PRs from new contributors, precisely because if you don't understand the code yourself and how to test it, you can't expect us to read your PRs and maintain your code. We get many PRs of this type, and it's becoming a huge problem.

This Repo is downstream of https://github.com/kiwix/kiwix-js/issues, and generally I suggest new contributors make their contributions upstream, taking into account the [Contribution Guide](https://github.com/kiwix/kiwix-js/blob/main/CONTRIBUTING.md), rather than here. We often onport contributions made upstream here. 

So, I strongly suggest you take a look at the upstream Repo, learn how to set up a proper testing instance, and find an issue there you would like to work on. The issue you chose is very complex, and unlikely to succeed unless you know the codebase here well. Closing here, but you are welcome to contribute upstream,

shashwat-28410:
Thankyou for the suggestion i will surely work on it !!

PR #933, opened 2026-08-21
Fix narrow-screen sizing classes for footer buttons and OPFS (#930)

### LINKED ISSUE #930: Narrow-screen sizing of the footer ToC and Top buttons is inert since the Bootstrap 4 migration
Found while sweeping for other Bootstrap 3 leftovers after #928, which had the same shape: a renamed class that silently stops matching, with no error to show for it.

### The problem

`resizeIFrame()` narrows the two footer buttons on small screens by swapping grid classes:

```js
if (window.outerWidth <= 470) {
    document.getElementById('dropup').classList.remove('col-xs-4');
    document.getElementById('dropup').classList.add('col-xs-3');
    if (window.outerWidth <= 360) {
        document.getElementById('btnTop').classList.remove('col-xs-2');
        document.getElementById('btnTop').classList.add('col-xs-1');
    } else {
        …
```

— `www/js/app.js`, around lines 161-174.

Bootstrap 4 dropped the `-xs` infix (`col-xs-4` became `col-4`), and the markup was migrated accordingly:

* `www/index.html:1538` — `<a … class="btn btn-outline-secondary dropdown-toggle col-4" id="dropup" …>`
* `www/index.html:1547` — `<a class="btn btn-outline-secondary col-2" id="btnTop" …>`

The JavaScript was not. So the `remove()` calls take away classes the elements never had, the `add()` calls add classes Bootstrap 4 does not define, and the `col-4` / `col-2` from the markup stay put. The ToC dropup and the Top button therefore keep their full width at every viewport size, and the narrow-screen adaptation has done nothing since `d7132d55` (2026-01-01), "Migrate project to Bootstrap 4 and miscellaneous cleanup/fixes #803 (#799)".

Nothing throws, and nothing looks broken unless you know the buttons are supposed to shrink — which is why it survived the migration testing.

### Fix

Drop the `-xs` from the eight class names, so the JavaScript toggles `col-4`/`col-3` on `#dropup` and `col-2`/`col-1` on `#btnTop`. Worth checking on a real narrow window afterwards that the intended proportions still look right under Bootstrap 4's grid, since the breakpoints (470px and 360px) predate it.

### Rest of the sweep was clean

For the record, so nobody repeats it — the other common Bootstrap 3 → 4 renames were checked across `www/index.html`, `www/js/` and `www/css/`:

* `pull-left` / `pull-right`, `img-responsive`, `hidden-xs` / `visible-xs`, `center-block`, `help-block`, `input-lg` / `input-sm`, `label-*` — no occurrences.
* `panel-heading` / `panel-body` — present only as element **ids** in `kiwixServe.js`; the classes there are already `card-header` and `card-body`.
* `btn-default` — no longer a Bootstrap class, but it is a real custom class in `www/css/app.css` (`.dark .btn-default`, `.darkfooter .btn-default`), so the toggling at `app.js:8163`/`:8169` is correct and should be left alone.
* `fade in` — fixed in #928; the only remaining occurrence in the tree is the word inside the DEV comment that documents that fix.

So `col-xs-*` looks like the last one of these.

bhuvan-somisetty:
Fixes #930

### Problem
Following the Bootstrap 4 migration, dynamic class manipulation in `www/js/app.js` still referenced Bootstrap 3 `col-xs-*` classes. Because Bootstrap 4 dropped the `-xs` infix, the `classList.remove()` and `classList.add()` calls were inert, preventing the Table of Contents dropup (`#dropup`) and Top button (`#btnTop`) from narrowing on small screens (<= 470px and <= 360px). Similar inert `col-xs-*` classes remained in the OPFS settings toggle.

### Solution
- Updated `resizeIFrame()` in `www/js/app.js` to toggle `col-4`/`col-3` on `#dropup` and `col-2`/`col-1` on `#btnTop`.
- Ensured `#btnTop` resets back to `col-2` when resizing to viewports > 470px.
- Updated `setOPFSUI()` to toggle `col-6`/`col-12` on `archiveFileLabel` and `col-5`/`col-6`/`col-7` on `archiveFileCol` and `archiveFilesCol`.

### Testing
- Ran ESLint on `www/js/app.js` (0 errors).
- Built unminified bundle via `npm run build-src` with Rollup/Babel.
- Swept repository to verify no remaining `col-xs-*` classes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Interim release 3.8.92
 
+* FIX: The ToC and Top buttons in the bottom navigation bar now resize correctly on narrow screens
+
 * FIX: Dropped ZIM files now load in contexts where the browser blocks the File System Access API, by falling back to the legacy file drop
 * FIX: Dropping a folder that the browser will not let the app read no longer voids the folder already picked
 * FIX: Links to articles that are HTML redirect stubs now open the target article correctly in Restricted mode
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -106,8 +106,11 @@ label {
 }
 
 #navigationButtons .dropup {
-    flex: 0 0 auto;
     position: relative;
+}
+
+#navigationButtons .dropup .btn {
+    width: 100%;
 }
 
 /* Position ToC dropdown above the button */
--- a/www/index.html
+++ b/www/index.html
@@ -1536,8 +1536,8 @@
                     <a class="btn btn-outline-secondary col-1" id="btnForward" title="Forward"><i class="fas fa-arrow-circle-right"></i></a>
                     <a class="btn btn-outline-secondary col-1" style="display:none;" id="btnRandomArticleAlt" title="Random Article"><i class="fas fa-random"></i></a>
                     <a class="btn btn-outline-secondary col-1" style="display:none;" id="btnToggleThemeAlt" title="Switch between light and dark modes"><i class="fas fa-adjust"></i></a>
-                    <div class="dropup">
-                        <a href="#" class="btn btn-outline-secondary dropdown-toggle col-4" role="button" id="dropup" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                    <div class="dropup col-4" id="dropupContainer">
+                        <a href="#" class="btn btn-outline-secondary dropdown-toggle" role="button" id="dropup" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                             <b>&nbsp;ToC&nbsp;</b>
                             <span class="caret"></span>
                         </a>
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -159,18 +159,20 @@ function resizeIFrame (reload) {
         ToCList.style.marginLeft = ~~(window.innerWidth / 2) - ~~(window.innerWidth * 0.16) + 'px';
     }
     if (window.outerWidth <= 470) {
-        document.getElementById('dropup').classList.remove('col-xs-4');
-        document.getElementById('dropup').classList.add('col-xs-3');
+        document.getElementById('dropup').classList.remove('col-4');
+        document.getElementById('dropup').classList.add('col-3');
         if (window.outerWidth <= 360) {
-            document.getElementById('btnTop').classList.remove('col-xs-2');
-            document.getElementById('btnTop').classList.add('col-xs-1');
+            document.getElementById('btnTop').classList.remove('col-2');
+            document.getElementById('btnTop').classList.add('col-1');
         } else {
-            document.getElementById('btnTop').classList.remove('col-xs-1');
-            document.getElementById('btnTop').classList.add('col-xs-2');
+            document.getElementById('btnTop').classList.remove('col-1');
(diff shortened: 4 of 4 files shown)


Jaifroid (changes requested):
@bhuvan-somisetty Half of this PR is right (the OPFS buttons, which are genuinely fixed). But the other half isn't working as mentioned in the other comment.

Here's what I'd suggest: open DevTools on the footer and compare the element the browser is actually laying out against the element carrying the col-* class -- and check whether toggling col-4 <-> col-3 by hand in the Elements panel changes the computed width at all. If hand-editing does nothing, no JS rename will either.

Also check the `#navigationButtons` block in www/css/app.css, added by the same Bootstrap 4 migration commit, to see if there is any grid layout at all.

Check the arithmetic across both footer configurations: Configuration -> navigation buttons at top reveals `#btnRandomArticleAlt` and `#btnToggleThemeAlt` in the footer (app.js:218-228), and `#lblZoom` sits in the row too. Do the columns sum to 12 in both states, at both breakpoints? There are various UI options that alter the number of buttons, or even move them from the bottom to the top. Those code patgs all need checking too.

When offering a fix and presenting test results, it would be helpful to provide screenshots at ~470px and ~360px in a genuinely narrow window, in both nav-button configurations.

Jaifroid (approved):
Thanks @bhuvan-somisetty -- this is exactly right now, and a much better fix than the first approach. Moving `col-4` onto the wrapper and dropping `flex: 0 0 auto` so the grid class supplies the flex basis gets at the actual cause. Thank you also for spotting that `#btnTop` was never reset to `col-2` on the way back up. That was a real latent bug.

I've tested visually at narrow widths in light and dark mode, with the navigation buttons in the bottom bar and switched to the top bar, and it all behaves. The column arithmetic holds in both configurations too: with the buttons at the top, `btnBack`/`btnForward` are hidden and the two Alt buttons shown, so the row still totals 12 either way.

(At 200% UI font size the ToC label clips, but that's inherent to the button now having a fixed width rather than growing, the toolbar icon remains usable, and any change there would belong in a separate PR. Nothing for you to do about it here.)

Two small things and then I'm happy to merge:

1. Please add a CHANGELOG entry under the `## Interim release 3.8.92` heading, along the lines of:
   `* FIX: The ToC and Top buttons in the bottom navigation bar now resize correctly on narrow screens`

2. The branch is one commit behind `main` (#941 landed since you opened this). I've test-merged it and it's clean, but please use "Update branch" so CI runs against current `main`. Remember to pull to your fork straight after you update the branch so that you don't have to force push anything

One optional extra, entirely up to you: while you were in this code you may have noticed `ToCList.style.marginLeft = ...` at `app.js:159` and `app.js:8274`. Both are dead: `#navigationButtons .dropup .dropdown-menu` in `app.css` sets `margin-left: 0 !important`, and an inline style assigned via `.style.marginLeft` can't carry `!important`, so the stylesheet wins in both places. It's another silently inert leftover from the same Bootstrap 4 migration as #930, so it would be fine in this PR if you'd like to remove those two lines (keep the adjacent `maxHeight` lines, as those do still do their job). Equally happy to take it as a follow-up.

When you've done the CHANGELOG and anything extra, let me know so we can run final tests and I can then squash/merge on your behalf.

PR #937, opened 2026-08-23
Fix brittle regex JSON parsing in updater.js (#936)

### LINKED ISSUE #936: Regex JSON parsing in updater.js is fragile to minified payloads
### Problem
In `www/js/lib/updater.js`, the in-app update checker (`getLatestUpdates()`) checks the GitHub Releases API (`https://api.github.com/repos/kiwix/kiwix-js-pwa/releases`) to notify users of new desktop releases (Electron, NWJS, UWP).

Release information is extracted via a global regular expression against the raw JSON response text:
```javascript
const regexpMatchGitHubReleases = RegExp('"browser_download_url[":\\s]+"(https:.*download\\/([^\\/]+).*(?:' + baseApp + ')[^"]+)"', 'ig');
```

**Note:** the GitHub API pretty-prints its JSON one field per line, and `.` doesn't match `\n`, so in practice the greedy `.*` below can't cross a field boundary against the live API and this hasn't caused any observed problem. This is a latent bug, not a live one — see discussion on #937.

This implementation has a latent parsing bug that would surface if the payload were ever minified or single-line:

1. **Greedy Wildcard (`.*`) Could Corrupt Asset URLs and Skip Releases on a Minified Payload**:
   The pattern `https:.*download\/` contains a greedy `.*`, which matches `"` as well as any other character. On a single-line JSON payload, this could match from the first `https:` across multiple release objects to the last `download/`, causing:
   - Intermediate/latest releases to be skipped.
   - Captured download URLs to contain raw JSON syntax from other release objects.
   - `tag` and `updateUrl` to be derived from the wrong release.

2. **State Leakage Across Repeated Update Checks via Top-Level Global RegExp**:
   `regexpMatchGitHubReleases` is defined as a module-level `const` with the `/g` flag, and is also built once at module-evaluation time from `params.packagedFile`, which may not be set yet.

3. **Fragility to Release Notes/Descriptions**:
   If a release body or commit message contains text mentioning `"browser_download_url"` immediately followed by a quoted string, the regex could in principle match against the release description rather than the structured asset metadata (this is separately guarded against because `[^"]` can't cross the description's own closing quote).

---

### Root Cause
Unanchored greedy wildcards (`.*`) inside the regex pattern that, given single-line/minified input, could match across JSON string boundaries. Additionally, the RegExp was built once at module scope with a `baseApp` value that could be stale, and carried `/g` state across calls.

---

### Affected Components
- `www/js/lib/updater.js` (`getReleasesObject`, `getLatestUpdates`)

---

### Acceptance Criteria
- Each regex capture group stays confined to the single JSON string value it started in, so a minified/single-line payload cannot cause cross-object matches.
- The RegExp (and the `baseApp` filter it depends on) is rebuilt fresh on every call, so there's no stale filter and no `lastIndex` state leakage between calls.
- Unit tests cover minified payloads, a realistic pretty-printed API response, channel variations, and malformed/empty input.

### Proposed Solution Direction
Replace each `.*` in the regex with `[^"]*` (or `[^"]+` where a non-empty match is required), since `[^"]` cannot match the quote that closes a JSON string, confining each match to one field. Build the RegExp inside `getLatestUpdates()` rather than at module scope.

bhuvan-somisetty:
Fixes #936

### Problem & Root Cause
In `www/js/lib/updater.js`, `getLatestUpdates()` checks the GitHub Releases API to notify desktop users (Electron, NWJS, UWP) of new releases, using a regex against the raw API text:
```javascript
const regexpMatchGitHubReleases = RegExp('"browser_download_url[":\\s]+"(https:.*download\\/([^\\/]+).*(?:' + baseApp + ')[^"]+)"', 'ig');
```

The greedy `.*` matches `"` as well as any other character, so on a minified/single-line payload it can match across multiple release/asset objects, corrupting the captured URL and skipping releases. The live GitHub API always pretty-prints its JSON one field per line, and `.` doesn't match `\n`, so this couldn't actually happen against production traffic — but it's still a latent bug worth closing, and the module-level `/g` RegExp also carried `lastIndex` state across calls and was built once from a `params.packagedFile` value that may not be set yet.

---

### Solution
Smallest fix that closes the gap, per review discussion:
1. Replace each `.*` in the regex with `[^"]*`/`[^"]+`, so a capture group can't cross the quote that closes the JSON string it started in — confines matching to one field even on minified input.
2. Build the RegExp fresh inside `getLatestUpdates()` (rather than once at module scope), using `getBaseAppPattern()` so it always reflects the current `params.packagedFile`, and avoiding `/g` state leakage between calls.
3. `getBaseAppPattern()` also fixes a real pre-existing bug: the old module-level `baseApp` constant read `params.packagedFile` before it was necessarily set.

No `JSON.parse` rewrite — kept the streaming regex read, both because the risk of a larger change isn't worth it for something that isn't reachable in the field, and because the response is fairly large (mostly a repeated GitHub user object per asset that's never used) and this update check also runs in the UWP build.

---

### Testing
- `test/updater.test.cjs`: minified payloads, channel matching, packaging filters, malformed/empty input, consecutive-call idempotency, a realistic pretty-printed fixture (`test/fixtures/github-releases-sample.json`) matching the real API's shape, and a live test that fetches the actual GitHub releases endpoint and skips gracefully if offline.
- `npm test`: all checks pass.
- `eslint`: 0 errors, 0 warnings on the changed files.
- `npm run build-src`: builds cleanly.

---

### DIFF
--- a/test/fixtures/github-releases-sample.json
+++ b/test/fixtures/github-releases-sample.json
@@ -1,0 +1,78 @@
+[
+  {
+    "url": "https://api.github.com/repos/kiwix/kiwix-js-pwa/releases/1",
+    "tag_name": "v4.0.0-E",
+    "target_commitish": "main",
+    "name": "v4.0.0-E",
+    "draft": false,
+    "prerelease": false,
+    "created_at": "2026-08-01T10:00:00Z",
+    "published_at": "2026-08-01T10:05:00Z",
+    "html_url": "https://github.com/kiwix/kiwix-js-pwa/releases/tag/v4.0.0-E",
+    "body": "See the changelog for details. Download links are also mirrored at https://download.kiwix.org/release/kiwix-js-pwa/ for convenience.",
+    "assets": [
+      {
+        "url": "https://api.github.com/repos/kiwix/kiwix-js-pwa/releases/assets/1001",
+        "id": 1001,
+        "name": "source.zip",
+        "content_type": "application/zip",
+        "state": "uploaded",
+        "size": 5242880,
+        "uploader": {
+          "login": "bhuvan-somisetty",
+          "id": 123456,
+          "avatar_url": "https://avatars.githubusercontent.com/u/123456?v=4",
+          "html_url": "https://github.com/bhuvan-somisetty",
+          "type": "User"
+        },
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/source.zip"
+      },
+      {
+        "url": "https://api.github.com/repos/kiwix/kiwix-js-pwa/releases/assets/1002",
+        "id": 1002,
+        "name": "kiwix-electron-4.0.0-E.exe",
+        "content_type": "application/x-msdownload",
+        "state": "uploaded",
+        "size": 104857600,
+        "uploader": {
+          "login": "bhuvan-somisetty",
+          "id": 123456,
+          "avatar_url": "https://avatars.githubusercontent.com/u/123456?v=4",
+          "html_url": "https://github.com/bhuvan-somisetty",
+          "type": "User"
+        },
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E.exe"
+      }
+    ]
+  },
+  {
+    "url": "https://api.github.com/repos/kiwix/kiwix-js-pwa/releases/2",
+    "tag_name": "v3.9.0-E",
+    "target_commitish": "main",
+    "name": "v3.9.0-E",
+    "draft": false,
+    "prerelease": false,
+    "created_at": "2026-05-01T10:00:00Z",
+    "published_at": "2026-05-01T10:05:00Z",
+    "html_url": "https://github.com/kiwix/kiwix-js-pwa/releases/tag/v3.9.0-E",
+    "body": "Minor fixes.",
+    "assets": [
+      {
+        "url": "https://api.github.com/repos/kiwix/kiwix-js-pwa/releases/assets/902",
+        "id": 902,
+        "name": "kiwix-electron-3.9.0-E.exe",
+        "content_type": "application/x-msdownload",
+        "state": "uploaded",
+        "size": 98765432,
+        "uploader": {
+          "login": "bhuvan-somisetty",
+          "id": 123456,
+          "avatar_url": "https://avatars.githubusercontent.com/u/123456?v=4",
+          "html_url": "https://github.com/bhuvan-somisetty",
+          "type": "User"
+        },
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v3.9.0-E/kiwix-electron-3.9.0-E.exe"
+      }
+    ]
+  }
+]
--- a/test/fixtures/github-releases-sample.json
+++ b/test/fixtures/github-releases-sample.json
@@ -30,8 +30,8 @@
       {
         "url": "https://api.github.com/repos/kiwix/kiwix-js-pwa/releases/assets/1002",
         "id": 1002,
-        "name": "kiwix-electron-4.0.0-E.exe",
-        "content_type": "application/x-msdownload",
+        "name": "kiwix-js-electron-4.0.0-E-x64.nsis.7z",
+        "content_type": "application/x-7z-compressed",
         "state": "uploaded",
         "size": 104857600,
         "uploader": {
@@ -41,15 +41,15 @@
           "html_url": "https://github.com/bhuvan-somisetty",
           "type": "User"
         },
-        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E.exe"
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-js-electron-4.0.0-E-x64.nsis.7z"
       },
       {
-        "name": "kiwix-electron-4.0.0-E-wikivoyage.exe",
-        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E-wikivoyage.exe"
+        "name": "kiwix-js-wikivoyage-4.0.0-E-arm64.nsis.7z",
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-js-wikivoyage-4.0.0-E-arm64.nsis.7z"
       },
       {
-        "name": "kiwix-electron-4.0.0-E-wikimed.exe",
-        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E-wikimed.exe"
+        "name": "kiwix-js-wikimed-4.0.0-E-arm64.nsis.7z",
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-js-wikimed-4.0.0-E-arm64.nsis.7z"
       }
     ]
   },
@@ -68,8 +68,8 @@
       {
         "url": "https://api.github.com/repos/kiwix/kiwix-js-pwa/releases/assets/902",
         "id": 902,
-        "name": "kiwix-electron-3.9.0-E.exe",
-        "content_type": "application/x-msdownload",
+        "name": "kiwix-js-electron-3.9.0-E-x64.nsis.7z",
+        "content_type": "application/x-7z-compressed",
         "state": "uploaded",
         "size": 98765432,
         "uploader": {
@@ -79,7 +79,7 @@
           "html_url": "https://github.com/bhuvan-somisetty",
           "type": "User"
         },
-        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v3.9.0-E/kiwix-electron-3.9.0-E.exe"
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v3.9.0-E/kiwix-js-electron-3.9.0-E-x64.nsis.7z"
       }
     ]
   }
--- a/test/fixtures/github-releases-sample.json
+++ b/test/fixtures/github-releases-sample.json
@@ -42,6 +42,14 @@
           "type": "User"
         },
         "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E.exe"
+      },
+      {
+        "name": "kiwix-electron-4.0.0-E-wikivoyage.exe",
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E-wikivoyage.exe"
+      },
+      {
+        "name": "kiwix-electron-4.0.0-E-wikimed.exe",
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E-wikimed.exe"
       }
     ]
   },
--- a/test/updater.test.cjs
+++ b/test/updater.test.cjs
@@ -65,21 +65,17 @@ function run (mockResponse, context) {
  * rejecting) if the API can't be reached, so a lack of network access skips the check instead
  * of failing the suite.
  *
- * GitHub varies the JSON formatting on the Accept header (note `Vary: Accept` on the response):
- * send one and you get pretty-printed JSON, one field per line; omit it and you get everything
- * minified onto a single line. https.get sends no Accept header by default, so with
- * withAcceptHeader left false this exercises the minified path - the shape that actually breaks
- * the old greedy-wildcard regex. Pass withAcceptHeader: true to fetch the pretty-printed shape
- * instead, which is what the app itself receives (uiUtil.XHR uses XMLHttpRequest, and browsers
- * add an `Accept: * / *` wildcard header to every request automatically).
+ * GitHub serves both a pretty-printed (one field per line) and a fully minified shape of the
+ * identical content, unpredictably, for identical requests a few minutes apart - it is not
+ * selected by the Accept header, HTTP version or User-Agent. So this fetches once and the
+ * caller derives the other shape locally with JSON.parse/JSON.stringify rather than relying on
+ * two live fetches to land on different shapes.
  *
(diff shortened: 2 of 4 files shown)


Jaifroid (changes requested):
Thanks for digging into this. The greedy .* in that regex is a genuine catch and worth removing.

One thing to flag on the framing, though: the GitHub API pretty-prints its JSON, one field per line, and `.` doesn't match a newline, so the wildcard can't actually escape the URL it started in. I checked the current live response against both versions of the code across all three packaging flavours and they agree on every case. The corruption you demonstrated is real, but it needs a minified payload, and I don't think anything in the chain produces one. So nothing was broken in the field, and I'd rather the issue and changelog didn't say it was. Could you soften the wording?

Given that, I'd prefer a smaller fix than moving to `JSON.parse`. Try `[^"]*` in place of each `.*` — since `[^"]` can't match the quote that ends a JSON string, the match is confined to one value, and it then works on minified input too.

I realise "don't use JSON.parse on JSON" sounds backwards, so let me say why I'd rather not here. As a default rule you're completely right, and if this were new code I'd probably write it your way. But we're not fixing a failure anyone is having: we're closing a hole that the server's output shape keeps shut anyway. When that's the situation, I want the smallest change that closes it, because the risk of a rewrite scales with its size and the benefit here is zero. That matters more than usual for this particular function: when it goes wrong it doesn't throw, it just quietly reports "Latest version" and nobody finds out for a release or two. A character-class change I can confirm by reading is a better trade than forty lines I have to trust the tests on.

There's a second reason to keep the streaming read. The releases response is around 1 MB, and over half of it is a GitHub user object repeated once per asset, which we never look at. Parsing it builds several hundred objects to extract thirty tags. That is neither here nor there in desktop Electron, but this update check also runs in the UWP build, which is still used on old Windows 10 Mobile (still supported), and on those the footprint is not free.

The other thing is that the rewrite brought changes I hadn't asked for: the tag now comes from `tag_name` rather than the asset URL, the link from `html_url`, and the flavour filter matches the tag as well as the filename. I checked and all three are harmless against today's releases, but they're behaviour changes travelling alongside a fix for something that wasn't happening, and that's the combination I'd rather avoid.

Two parts of your PR I'd like to keep either way: `getBaseAppPattern()` (the old module-level constant read `params.packagedFile` before it was set -- a real bug you've caught) and building the RegExp inside the function.

The tests are a welcome addition. One suggestion: every fixture is hand-written and minified, which is why the original code looks broken under them. A capture of a real API response would be worth adding: it's a shape that's hard to invent, and it would have shown up the difference straight away. So the test should check the API is available and check code against what's actually coming from it.

Last, a process point, and I'm not raising this to put you off. CONTRIBUTING asks that you open an issue and agree the approach before writing code, precisely so that effort isn't spent on something I then push back on (which is what has happened here). The in-app updater is squarely this repo's territory, so this one is filed in the right place, but at least one of your recent issues touches shared app code, and those should be filed first upstream in kiwix-js, where a fix would benefit the browser extensions too. Please see CONTRIBUTING on this point.

And please hold issues to the same standard as PRs. This one reported a live fault with corrupted URLs and skipped releases, when a single check against the real API response would have shown none of it was reachable. I'm an unpaid volunteer doing this in my spare time, and a confidently worded report takes an evening of my time to disprove.

www/js/lib/updater.js
```diff
@@ -35,12 +35,11 @@ params.updateServer = {
 };
 
 // A RegExp prototype string to match the current app's releases
```
Jaifroid:
```suggestion
/**
 * Builds the RegExp prototype string that matches the current app's releases. Evaluated on each
 * check rather than once at import time, because params.packagedFile is not necessarily populated
 * when this module is first loaded.
 *
 * @returns {String} An alternation of filename fragments identifying this app's own packages
 */
```
The old one-line comment described a constant; this is a function now, and the repo puts JSDoc on functions. The reason for that (that it must be re-evaluated per call) is currently ~45 lines away inside `getLatestUpdates`, so most readers won't see it.

www/js/lib/updater.js
```diff
@@ -78,6 +77,12 @@ function getLatestUpdates (callback) {
         var releaseFile;
         var releaseVersion;
         var releaseChannel;
+        // Build the RegExp fresh on every call: baseApp depends on params.packagedFile, which
+        // may not be set yet when this module is first evaluated, and a module-level /g RegExp
+        // would also carry its lastIndex over between calls.
+        // [^"]+ is used in place of a plain wildcard so each capture group stays confined to the
+        // single JSON string value it started in, even if the payload is minified to one line.
```
Jaifroid: Small thing: the pattern uses `[^"]*` in two of the three positions, and the comment says `[^"]+`. Could you also spell out why it works? The mechanism is the important part -- `[^"]` cannot match the quote that terminates a JSON string, so a match physically cannot run into the next field. Without that, someone might eventually "tidy" this back to `.*`.

www/js/lib/updater.js
```diff
@@ -78,6 +77,12 @@ function getLatestUpdates (callback) {
...
+        var regexpMatchGitHubReleases = RegExp('"browser_download_url[":\\s]+"(https:[^"]*download\\/([^\\/"]+)[^"]*(?:' + getBaseAppPattern() + ')[^"]+)"', 'ig');
         // Loop through every line in releases
```
Jaifroid: "Loop through every matching asset URL" would be more accurate now.

test/updater.test.cjs
```diff
@@ -0,0 +1,229 @@
...
+ * of failing the suite.
+ *
+ * @returns {Promise<String|null>} The raw response text, or null if it could not be fetched
+ */
+function fetchLiveReleases () {
+    return new Promise(function (resolve) {
+        var req = https.get('https://api.github.com/repos/kiwix/kiwix-js-pwa/releases', {
+            headers: { 'User-Agent': 'kiwix-js-pwa-updater-test' },
+            timeout: 5000
+        }, function (res) {
```
Jaifroid, line 82: Worth knowing: this helper is exercising the minified path, not the pretty-printed one. GitHub varies its formatting on the Accept header (note `Vary: Accept` on the response). If you send one, you get one field per line; omit it, and you get everything on a single line. `https.get` doesn't send one by default, so this fetch gets minified JSON. That's why this check fails against the old regex, and it makes it the strongest test here, but that's by accident rather than design, and nothing records that.

Two things to do:

* Please note this in the JSDoc above, so nobody later "fixes" it by adding an Accept header and silently neuters the check.
* Better still, fetch it both ways -- once with Accept: */* and once without -- and assert the two produce identical results. That would be a direct test of the invariant this PR establishes, on real data, and it needs about three extra lines.

(This also corrects something I told you earlier: I said the API always pretty-prints. I double-checked, and it doesn't: it depends on the request. The app is fine because `XMLHttpRequest` always sends `Accept: */*`, but that's a browser default we never set explicitly, which makes this fix better justified than "latent".)

test/updater.test.cjs
```diff
@@ -0,0 +1,229 @@
...
+            if (res.statusCode !== 200) {
+                res.resume();
+                resolve(null);
+                return;
+            }
+            var data = '';
+            res.on('data', function (chunk) { data += chunk; });
```
Jaifroid, line 90: `data += chunk` stringifies each Buffer separately, so a multi-byte UTF-8 character split across a chunk boundary gets mangled. Today's live response has a handful of them, so this could flake depending on where the TCP chunks land. You should use `res.setEncoding('utf8')` before the handler, which makes Node buffer the partial sequence for you.

test/updater.test.cjs
```diff
@@ -0,0 +1,229 @@
...
+
+    const malformedJson = await run('not valid json {[[', { appVersion: '3.8.0' });
+    check('Handles malformed JSON gracefully without throwing', malformedJson.updateTag === undefined && malformedJson.updatedReleases.length === 0);
+
+    const emptyResponse = await run('', { appVersion: '3.8.0' });
+    check('Handles empty response gracefully', emptyResponse.updateTag === undefined && emptyResponse.updatedReleases.length === 0);
+
+    section('Real API response shape (pretty-printed, one field per line)');
+    // GitHub's REST API pretty-prints its JSON responses with one field per line, which is
+    // what actually reaches this code in production - unlike the hand-minified fixtures above.
```
Jaifroid: This says the API pretty-prints "which is what actually reaches this code in production". The conclusion is right but the reason isn't -- as above, the format depends on the `Accept` header. The accurate version is that the API pretty-prints for clients that send `Accept`, and `uiUtil.XHR` gets that formatting because browsers add `Accept: */*` automatically. Worth stating precisely, since it's exactly the assumption this code rests on.

test/updater.test.cjs
```diff
@@ -0,0 +1,229 @@
...
+    if (liveReleasesText === null) {
+        console.log('  skip  Could not reach the GitHub releases API - skipping live comparison');
+    } else {
+        let liveRes;
+        let threw = false;
+        try {
+            liveRes = await run(liveReleasesText, { appVersion: '0.0.1' });
+        } catch (e) {
+            threw = true;
+        }
```
Jaifroid, line 246: Given this is the only check running against real data, the assertions are doing less work than they could. `appVersion: '0.0.1'` means everything matches, and we only verify the URLs are well-formed -- a regression that returned the wrong release would pass. Could you assert the detected tag against the newest matching release from the parsed payload, so it checks the answer and not just the format?

test/fixtures/github-releases-sample.json
Jaifroid, line 1: Thanks for adding the surrounding noise fields -- that's a real improvement on the earlier fixtures. It is still invented data though: two releases, no WikiMed/Wikivoyage interleaving, and a future-dated v4.0.0-E. A saved capture of an actual response would cover the flavour filter against tags and filenames we really ship, and it's a shape that's hard to invent convincingly. Trimming it to the four fields the parser reads keeps it small. You should add a Wikivoyage and/or WikiMed release too.

Jaifroid (changes requested):
Thank you very much for this -- it's a great improvement! And it turns out it fixes a real potential weakness too (see the comments below about the effect of the `Accept` header). There are a few further nitpicks and also some ways in which the tests could be made more robust. Please take a look.

test/updater.test.cjs
```diff
@@ -0,0 +1,267 @@
...
+        // The Accept-header behaviour documented on fetchLiveReleases is exactly the assumption
+        // this fix rests on: assert the minified and pretty-printed shapes of the same live data
+        // produce identical results, as a direct test of that invariant
+        const liveReleasesTextWithAccept = await fetchLiveReleases(true);
+        if (liveReleasesTextWithAccept !== null) {
+            const liveResWithAccept = await run(liveReleasesTextWithAccept, { appVersion: '0.0.1' });
+            check('Minified (no Accept header) and pretty-printed (Accept: */*) responses agree',
+                liveResWithAccept.updateTag === liveRes.updateTag &&
+                JSON.stringify(liveResWithAccept.updatedReleases) === JSON.stringify(liveRes.updatedReleases));
+        }
```
Jaifroid: This is comparing the same shape against itself. Both fetches returned byte-identical minified payloads when I ran it (878,532 bytes each), so the check reported success without ever comparing a pretty-printed response to a minified one.

The premise it's built on is also wrong, and that's my fault rather than yours: I incorrectly told you the `Accept` header selects the format. On further testing, it turns out GitHub serves both shapes for identical requests: the same URL, the same headers, same client, just a few minutes apart. I've since ruled out `Accept`, HTTP version and User-Agent as predictors. The two bodies are the same content; the minified one is exactly `JSON.stringify` of the pretty one.

So please don't depend on the network to hand you both shapes. Fetch once and derive the second shape locally. Build both shapes locally from whatever arrives, and assert they really are different before comparing:

```
const parsed = JSON.parse(responseText);
const asMinified = JSON.stringify(parsed);              // byte-identical to GitHub's minified form
const asPretty = JSON.stringify(parsed, null, 2) + '\n'; // byte-identical to GitHub's pretty form
```

I checked both against real responses: `JSON.stringify(parsed)` matches GitHub's minified body exactly, and `JSON.stringify(parsed, null, 2)` matches its pretty body exactly bar the trailing newline. Assert `asPretty` has many lines and `asMinified` has one, then run the parser over each and compare the results. That can't pass vacuously, needs one fetch instead of two, and removes the race between them. So `withAcceptHeader` and its `JSDoc` can go, with my apologies for having given you wrong info. 🙏🙏

It's also worth doing this against the checked-in fixture as well as the live response: the invariant is then tested even offline, which is where it matters most since the live check skips whenever the network is unavailable.

test/fixtures/github-releases-sample.json
```diff
@@ -0,0 +1,86 @@
...
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E.exe"
+      },
+      {
+        "name": "kiwix-electron-4.0.0-E-wikivoyage.exe",
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E-wikivoyage.exe"
+      },
+      {
+        "name": "kiwix-electron-4.0.0-E-wikimed.exe",
+        "browser_download_url": "https://github.com/kiwix/kiwix-js-pwa/releases/download/v4.0.0-E/kiwix-electron-4.0.0-E-wikimed.exe"
+      }
```
Jaifroid, line 53: Two things here. First, nothing exercises these: the fixture is read once at line 205 and run with `packagedFile` unset, while the flavour checks at lines 185-188 use their own inline payload. As it stands the addition is inert.

Second, the names don't match what we actually publish. Real flavour assets look like `kiwix-js-wikivoyage-3.8.2-E-arm64.nsis.7z` and `kiwix-js-wikimed-3.7.6-E-arm64.nsis.7z` — note there's no `electron` in them. I checked every asset across the last 30 releases and none matches both a flavour name and the default `windows|electron|kiwixwebapp_` pattern. The incorrect `kiwix-electron-4.0.0-E-wikivoyage.exe` matches both, so it models a convention that doesn't exist and would make a default-build check pick up flavour assets, which never happens in reality.

Could you rename them to the real convention and add a check that runs this fixture with `packagedFile: 'wikivoyage_en_all_maxi.zim'` (and the WikiMed equivalent), asserting only the flavour assets come back?

Jaifroid (changes requested):
@bhuvan-somisetty Thanks for turning this round: you've picked up every one of the review points cleanly, and the code is noticeably better for it. The JSDoc, the corrected comments, the `setEncoding` fix and the stronger live assertions are all exactly what was wanted.

I owe you an apology: it turns out your original diagnosis was right, and my pushback was wrong. I've since established that the API does sometimes serve minified JSON -- in one 13-minute window I got it on every single request, whereas earlier this week I was getting pretty printed -- so this was a live bug, not a latent one.

We're nearly there. Just a couple of small things below, and I think this will then be good to merge.

Jaifroid (approved):
Thank you @bhuvan-somisetty , your latest commit addresses all issues. I'll proceed to squash-merge. Thank you for a valuable PR that meaningfully patches and improves the PWA's update handling.

Issue #956, opened 2026-09-15
Make exit full screen icon/button into an actual toggle


Jaifroid:
Currently the exit full screen icon turns off the whole full-screen system, and the only way to get back into it is via the dropdown in configuration. If the user clicks the button, then the auto-reactivation should be suppressed until the user toggles it back on, with the full-screen icon remaining in place.

We should probably also monitor the other methods of exiting full screen (Esc or F11, at least on Windows), and ensure these have consistent behaviour. The auto-restore exists because there are several actions in the app that auto-exit full screen, especially on mobile, and we need to restore with a user gesture. We need to distinguish somehow between intended exiting and accidental exiting, and this needs more careful consideration.

PR #962, opened 2026-09-15
Avoid exitFullscreen rejection when cancelling in windowed mode #961

### LINKED ISSUE #961: requestOrCancelFullScreen unconditionally invokes exitFullscreen when not in fullscreen, causing unhandled rejection
`requestOrCancelFullScreen()` in `www/js/lib/uiUtil.js` unconditionally calls `document.exitFullscreen()` when called without arguments, even if the app isn't currently in fullscreen mode.

Per the W3C Fullscreen API spec, calling `document.exitFullscreen()` when `document.fullscreenElement === null` rejects with a `TypeError` ("Document not active"). `requestOrCancelFullScreen` already guards entering fullscreen (`if (el && appIsFullScreen()) return Promise.resolve(true);`), but has no symmetric guard for `!el && !appIsFullScreen()`.

## Steps to Reproduce

1. Open the app in normal windowed mode.
2. Go to **Configuration** -> **Display & Presentation** -> **Lock screen orientation**, and choose an orientation option (any non-empty option enters fullscreen; orientation options generally only work in mobile browsers, but any option still triggers the fullscreen request).
3. Press **Esc** to leave fullscreen (the app's own state doesn't know this happened yet).
4. Set the dropdown back to **Normal** (`''`). This calls `uiUtil.lockDisplayOrientation('')`, which calls `requestOrCancelFullScreen()` with no argument, which calls `document.exitFullscreen()` while the document is already not fullscreen.

## Root Cause

```javascript
function requestOrCancelFullScreen (el) {
    // Don't do anything if already in full-screen mode, and user has not requested to exit full-screen mode
    if (el && appIsFullScreen()) {
        console.debug('Display is already full screen');
        return Promise.resolve(true);
    }
    // Choose the correct method to request or cancel full-screen mode
    var rq = function (sel) {
        var fn = sel
            ? sel.requestFullscreen ? sel.requestFullscreen()
            : ...
            : document.exitFullscreen ? document.exitFullscreen()
            : ...;
        return Promise.resolve(fn);
    };
    return rq(el).then(function () {
        console.log(el ? 'Full-screen mode enabled' : 'Full-screen mode disabled');
        return !!el;
    }).catch(function (err) {
        console.log('Error enabling full-screen mode', err);
        throw err;
    });
}
```

## Impact

The dropdown's `change` handler in app.js already has a `.catch` around this call, so the rejection is caught, not unhandled — the visible result is a logged `Error locking screen orientation`, not an unhandled promise rejection. But that `.catch` doesn't call `setDynamicIcons()` the way the success path does:

```javascript
uiUtil.lockDisplayOrientation().then(function () {
    setDynamicIcons();
}).catch(function () {
    console.log('Error locking screen orientation');
});
```

So after following the steps above, the navbar keeps showing the "Return to fullscreen" (expand) icon even though the dropdown now says Normal and the app is windowed — this is the actual symptom a user sees. (`refreshFullScreen`'s own exit-fullscreen call in app.js is not affected, since it only runs when `appIsFullScreen()` is already true.)

## Acceptance Criteria

1. Calling `requestOrCancelFullScreen()` when already in windowed mode resolves to `false` without calling `document.exitFullscreen()` or throwing an error.
2. Setting the orientation lock dropdown back to Normal after manually leaving fullscreen no longer logs an error, and the navbar icon updates correctly.
3. Automated regression tests verify that cancelling fullscreen when not fullscreen resolves cleanly without throwing.

## Solution Direction

Add the symmetric early return in `requestOrCancelFullScreen()`:

```javascript
function requestOrCancelFullScreen (el) {
    // Don't do anything if already in full-screen mode, and user has not requested to exit full-screen mode
    if (el && appIsFullScreen()) {
        console.debug('Display is already full screen');
        return Promise.resolve(true);
    }
    // Don't do anything if already not in full-screen mode, and user requested to exit full-screen mode
    if (!el && !appIsFullScreen()) {
        console.debug('Display is already not full screen');
        return Promise.resolve(false);
    }
```

bhuvan-somisetty:
Fixes #961

### Problem & Root Cause

In `www/js/lib/uiUtil.js`, `requestOrCancelFullScreen(el)` requests full-screen mode when `el` is provided, and cancels full-screen mode when `el` is omitted / falsy.

While the function already checked if the display is already full screen when requesting entry:
```javascript
if (el && appIsFullScreen()) {
    console.debug('Display is already full screen');
    return Promise.resolve(true);
}
```
it lacked the corresponding check for cancellation. When called with no arguments while the app is in normal windowed mode, it proceeded to invoke `document.exitFullscreen()`.

According to the W3C Fullscreen API specification, calling `document.exitFullscreen()` when `document.fullscreenElement === null` rejects with a `TypeError` (`"Document not active"`).

To reproduce: choose an orientation option in Configuration -> Display & Presentation (any non-empty option enters fullscreen), press Esc to leave fullscreen, then set the dropdown back to Normal. That calls `uiUtil.lockDisplayOrientation('')` -> `requestOrCancelFullScreen()` with no argument, which calls `document.exitFullscreen()` while already windowed.

### Impact

The dropdown's `change` handler in app.js already wraps this call in a `.catch`, so the rejection is caught rather than unhandled — the visible result is a logged `Error locking screen orientation`, not an unhandled promise rejection. But that `.catch` doesn't call `setDynamicIcons()` the way the success path does, so the navbar keeps showing the "Return to fullscreen" (expand) icon after the dropdown is set back to Normal — that's the symptom a user actually sees. This fix also resolves that, because the exit call now resolves instead of rejecting, so the `.then` (and its `setDynamicIcons()` call) runs.

### Solution

1. **Early Return on Redundant Cancellation**: Added a symmetric guard in `requestOrCancelFullScreen(el)`:
   ```javascript
   if (!el && !appIsFullScreen()) {
       console.debug('Display is already not full screen');
       return Promise.resolve(false);
   }
   ```
2. **Clarified Catch Logging**: Updated catch block log message from `'Error enabling full-screen mode'` to accurately distinguish between enabling and disabling failures.
3. **Automated Unit Tests**: Added `test/fullscreen.test.cjs` and registered it in `package.json` under `npm test`:
   - Cancellation in windowed mode returns `false` without calling `exitFullscreen()` and does not throw.
   - Orientation reset (`lockDisplayOrientation('')`) in windowed mode resolves cleanly.
   - Normal entry and exit flows, including re-request idempotency.
   - Verified rejection propagation when `exitFullscreen()` fails in active fullscreen mode.
4. **CHANGELOG**: Added an entry describing the fix.

(No new export was added to `uiUtil.js` — the test evaluates the source directly, and app.js doesn't need to call `requestOrCancelFullScreen` itself, so the module's public API is unchanged, leaving that for the fullscreen rework in #956 to decide.)

---

### Testing
- `npm test`: All 97 checks passed across all 4 suites (`overrideParams.test.cjs`, `updater.test.cjs`, `extractImagesMedia.test.cjs`, `fullscreen.test.cjs`).
- `node ./node_modules/eslint/bin/eslint.js www/js/lib/uiUtil.js test/fullscreen.test.cjs`: 0 errors.
- `npm run build-src`: Bundled successfully.

---


Jaifroid (changes requested):
@bhuvan-somisetty thanks for this. I've checked it: the bug is real, and the guard is a good fix. I tested it in Chromium 150, and calling `document.exitFullscreen()` when the window isn't fullscreen does reject with `TypeError: ... Document not active`. The new early return mirrors the existing check for entering fullscreen, and no caller uses the return value, so it's safe.

Before I merge, could you fix a few things in the code and in the issue/PR description?

**Code**

1. Please put back the UTF-8 BOM in `www/js/lib/uiUtil.js`. The diff removes it on line 1 (`-﻿/**` -> `+/**`). Probably your editor did this. Every other file in `www/js/lib/` and `www/js/app.js` has a BOM (the Microsoft Store requires the BOM), so please keep this one the same and leave that line out of the diff.
2. Please remove the new `requestOrCancelFullScreen` export. Nothing uses it: your test evals the source instead of importing the module, and app.js doesn't call it. We're planning to rework the fullscreen button in #956, and that work will decide what app.js needs, so I'd rather not add to the module's API now.

**Issue #961 and the PR description**

These points don't affect the fix, but the issue will be read later as a record of the bug, so please correct them:

1. "Unhandled promise rejection": this doesn't happen in the current code. The only exit call without a `.catch` is the `btnAbout` branch of `refreshFullScreen` in app.js, and it only runs when `appIsFullScreen()` is already true. The path that actually fails is setting the Configuration dropdown to "Normal", and its `.catch` handles the error. The real result is a logged error, not an unhandled rejection.
2. "Select any option" in the steps to reproduce is wrong. Any non-empty option enters fullscreen. Only "Normal" ('') exits, so only that option reproduces the bug. To reproduce it reliably: choose an orientation option so the app goes fullscreen, press Esc to leave fullscreen, then set the dropdown back to "Normal" (orientation options only usually work in mobile browsers).
3. "On initial reset": I can't find a caller that exits fullscreen at startup or on reset. If you meant a specific path, please name it. If not, please remove it.
4. You missed a user-visible symptom. In the dropdown handler, the `.catch` doesn't call `setDynamicIcons()`. So with the steps in point 2 above, the navbar keeps showing the "Return to fullscreen" (expand) icon after you've set the option to Normal. Your fix also fixes this, because the exit call now resolves and the `.then` runs. It's worth adding to the Impact section, since it's the only effect a user actually sees.

One general request: please check claims like "unhandled rejection" and the reproduction steps against the running app before filing. It makes review much quicker.

Thanks again, the fix itself is good.

Jaifroid:
PS Please update the branch, and pull the change to your fork, as usual.

Jaifroid:
I've just remembered we also need a brief CHANGELOG entry.

bhuvan-somisetty:
Thanks for the thorough review — all fixed:

- Restored the BOM in uiUtil.js (yep, my editor stripped it).
- Removed the unused `requestOrCancelFullScreen` export, leaving that decision to #956.
- Rebased onto latest main and force-pushed.
- Added a CHANGELOG entry.
- Corrected the issue and PR description: it's a caught error (not unhandled), fixed the reproduction steps to the actual "orientation option -> Esc -> Normal" path, dropped the unverifiable "on initial reset" claim, and added the real user-visible symptom — the navbar icon staying on "Return to fullscreen" after setting Normal, since the dropdown's `.catch` skipped `setDynamicIcons()`.

All 97 checks pass, eslint clean. Thanks for pushing me to verify the actual behaviour instead of assuming it — makes sense now.

bhuvan-somisetty:
Rebased onto latest main and resolved the CHANGELOG conflict (just two entries landing in the same spot) — force-pushed, all still green.

Jaifroid (approved):
Great, thanks for this PR. It's good to go, so I'll squash/merge.

PR #969, opened 2026-09-18
Run the unit tests in CI, and share the test harness between suites (#968)

### LINKED ISSUE #968: Run the unit tests in CI, and share the test harness between suites
We now have four unit test suites under `test/`, but none of the GitHub workflows actually runs them, so `npm test` only gets run when someone remembers to do it locally. It would be good to have a small workflow that runs `npm ci && npm test` on pushes and pull requests, so that a regression shows up on the PR itself.

While we're at it, the four suites each re-implement the same scaffolding: reading the module source and stripping its import/export lines, evaluating it inside a function with mocked globals, a hand-rolled assert or check with pass/fail counters, and the summary line with the exit code. That's around 50 lines per file before any actual test is written, which is why the recent fullscreen test came out at over 200 lines for a five-line fix. It also means `npm test` in package.json is a growing chain of &&s that has to be edited every time a suite is added.

I think the cleanest way to reduce this is to move to Node's built-in test runner (`node --test`, available since Node 18). It gives us `describe/it`, `node:assert`, automatic discovery of *.test.cjs files, a single aggregated result, and TAP or JUnit output that the workflow can read. That would remove all the hand-rolled counters and exit-code handling. The source-loading and eval-with-globals part could then go into a small shared `test/helpers.cjs`, perhaps with a minimal mock document factory for the bits that are the same everywhere (`getElementById` returning a stub, `querySelectorAll` returning an empty array, and so on). The mocks that give each test its meaning would stay in the individual files.

I'd rather not add jsdom or a third-party test framework for this. The suites are small, jsdom doesn't implement things like the Fullscreen API anyway, and the built-in runner covers what we need without a new dependency.

The longer-term problem, that we have to strip imports and eval the source at all because the modules import each other and touch `document` and `localStorage` at load time, is a separate and bigger job that could be tackled here or in a separate issue/PR.

bhuvan-somisetty:
## Summary

- Adds `.github/workflows/run-tests.yml`, running `npm ci && npm test` on every push to `main` and on every pull request, so a regression now shows up on the PR itself instead of relying on someone remembering to run the suites locally.
- Migrates all four suites under `test/` from hand-rolled pass/fail counters, `console.log`-based reporting and `process.exit` handling to Node's built-in test runner (`node --test`), using `describe`/`it` and `node:assert`. `npm test` is now `node --test test/*.test.cjs`, which discovers and aggregates all suites, so it no longer needs editing every time a suite is added.
- Factors the scaffolding that was duplicated in every suite (stripping ES module import/export lines before evaluating the source in a sandbox, and a small mock `document` with the same `getElementById`/`querySelector`/`querySelectorAll`/`createElement` stubs) into a new `test/helpers.cjs`. The mocks that give each test its actual meaning, and the eval-with-globals sandbox each suite builds, stay in the individual files since they're specific to the module under test.
- No jsdom or third-party test framework added, per the issue - the built-in runner covers `describe`/`it`, `node:assert`, automatic `*.test.cjs` discovery, and a single aggregated pass/fail result.

The bigger, separate problem noted in the issue - that the suites have to strip imports and eval the source at all because the app's modules import each other and touch `document`/`localStorage` at load time - is left for a follow-up.

## Test plan

- [x] `npm test` passes locally: 72 checks across 4 suites (`overrideParams`, `updater`, `extractImagesMedia`, `fullscreen`)
- [x] `npx eslint --ext .cjs test` passes clean
- [x] Verified `npm test` works from both a POSIX shell and native PowerShell on Windows (the plain directory form `node --test test/` has a resolution bug on this Node 24/Windows combination, so the script uses the explicit glob `test/*.test.cjs` instead)


Jaifroid (changes requested):
Thanks, this is a clean migration and does what the issue asked. All 72 checks pass for me locally on Node 22, and eslint is clean. Two small requests:

1. Collapse the order-dependent tests in test/fullscreen.test.cjs. In the first two describe blocks ("Cancelling full-screen mode when already in windowed mode" and "Resetting display orientation lock in windowed mode"), the first `it` performs the action and stores the outcome in a closure variable (`result`, `threw`), and the following it blocks assert on that variable. That only works because the runner happens to execute them in order; under `--test-only` or `--test-name-pattern` filtering the later checks would run against unset state. Please fold each of those two blocks into a single `it` that awaits the call and then makes all the assertions. It'll also trim a few lines. The other `describe` blocks are fine as they are.

2. Document the Node version floor. Because npm on Windows doesn't expand `test/*.test.cjs`, the glob is passed to Node literally, and `node --test` only started expanding globs itself in Node 21. So on Windows `npm test` needs Node 21 or later, while on Linux/macOS (and in CI) the shell expands it and Node 20 is enough. Could you note that in two places:

* CONTRIBUTING.md, the paragraph beginning "There is one important difference from upstream" (currently line 35). It still says this repo has no unit test suite and that CI only checks the build, which your PR makes untrue. Please rewrite it to say we now have unit tests under test/, that CI runs them on every PR, that they can be run locally with `npm test`, and the Node version requirement above. Keep the point that the suites are small and that manual testing is still required.
* The header comment of `test/helpers.cjs`, one sentence with the same Node requirement, since that's the file anyone adding a suite will open.

But this is looking very good and a great addition to the Repo.

bhuvan-somisetty:
Thanks for the review. Both done.

Collapsed those two fullscreen.test.cjs blocks into single it calls each, so they don't lean on execution order anymore. Also added the Node version note in CONTRIBUTING.md and in the helpers.cjs header, explaining why Windows needs Node 21+ while Linux/macOS only needs 20+.

68 checks pass locally now (down from 72 since the collapsed blocks merged a few asserts), eslint still clean.

Jaifroid (approved):
All looks good now, thank you very much!

PR #971, opened 2026-09-19
On-port: encode url and title in DirEntry toStringId (#966)

### LINKED ISSUE #966: DirEntry.toStringId does not escape pipe delimiter, corrupting title, redirect and navigation on entries containing pipe character
When articles are listed in the search autocomplete dropdown or the ZIM Index modal, their directory entries are serialized into the `dirEntryId` HTML attribute via `dirEntry.toStringId()` (`app.js:5463`, `app.js:5672`). When the user clicks an item, `findDirEntryFromDirEntryIdAndLaunchArticleRead()` deserializes it back into a `DirEntry` via `DirEntry.fromStringId()` (`app.js:5736`).

`DirEntry.prototype.toStringId` constructs a pipe-delimited string of 9 fields:
```javascript
DirEntry.prototype.toStringId = function () {
    return this.offset + '|' + this.mimetypeInteger + '|' + this.namespace + '|' + this.cluster + '|' +
            this.blob + '|' + this.url + '|' + this.title + '|' + this.redirect + '|' + this.redirectTarget;
};
```

Neither `this.url` nor `this.title` is escaped or encoded when building this string. When `DirEntry.fromStringId()` deserializes the string using `stringId.split('|')`, any pipe character (`|`) present within the title or URL introduces extra tokens into `idParts`.

This causes all subsequent fields to be read from the wrong index:
- `title` is truncated at the first `|`.
- `redirect` (index 7) receives the remainder of the title instead of the redirect boolean (`"..." === 'true'` evaluates to `false`).
- `redirectTarget` (index 8) receives the boolean string `"true"` or `"false"` instead of the target URL index.
- If the entry was a redirect, `dirEntry.isRedirect()` evaluates to `false`, causing the click handler to skip `resolveRedirect()` and attempt to read invalid cluster/blob offsets directly.
- If `url` contains a pipe character, both `url` and `title` are misaligned and corrupted, breaking navigation completely.

## Steps to Reproduce

Run the following test in Node:
```javascript
import zimDirEntry from './www/js/lib/zimDirEntry.js';
const DirEntry = zimDirEntry.DirEntry;

const fakeZimFile = {
    mimeTypes: new Map([[1, 'text/html']]),
    blob: () => Promise.resolve(new Uint8Array())
};

// Article title with a pipe character (common in programming, math, logic, and wiki articles)
const original = new DirEntry(fakeZimFile, {
    offset: 1000,
    mimetypeInteger: 1,
    namespace: 'C',
    cluster: 42,
    blob: 7,
    url: 'Bitwise_OR_operator',
    title: 'Bitwise | (OR) operator in C/C++',
    redirect: true,
    redirectTarget: 8888
});

const serialized = original.toStringId();
const reconstructed = DirEntry.fromStringId(fakeZimFile, serialized);

console.log('Original title:', original.title);
console.log('Reconstructed title:', reconstructed.title);
console.log('Original redirect:', original.redirect);
console.log('Reconstructed redirect:', reconstructed.redirect);
console.log('Original redirectTarget:', original.redirectTarget);
console.log('Reconstructed redirectTarget:', reconstructed.redirectTarget);
```

Output:
```text
Original title: Bitwise | (OR) operator in C/C++
Reconstructed title: Bitwise 
Original redirect: true
Reconstructed redirect: false
Original redirectTarget: 8888
Reconstructed redirectTarget: true
```

## Root Cause

In `www/js/lib/zimDirEntry.js`:
```javascript
DirEntry.prototype.toStringId = function () {
    return this.offset + '|' + this.mimetypeInteger + '|' + this.namespace + '|' + this.cluster + '|' +
            this.blob + '|' + this.url + '|' + this.title + '|' + this.redirect + '|' + this.redirectTarget;
};

DirEntry.fromStringId = function (zimfile, stringId) {
    var data = {};
    var idParts = stringId.split('|');
    data.offset = parseInt(idParts[0], 10);
    data.mimetypeInteger = parseInt(idParts[1], 10);
    data.namespace = idParts[2];
    data.cluster = parseInt(idParts[3], 10);
    data.blob = parseInt(idParts[4], 10);
    data.url = idParts[5];
    data.title = idParts[6];
    data.redirect = (idParts[7] === 'true');
    data.redirectTarget = idParts[8];
    data.fromArticleList = true;
    return new DirEntry(zimfile, data);
};
```

Using naive `split('|')` on unescaped user-supplied title and URL strings breaks whenever those strings contain the delimiter.

## Impact

- Clicking search results or ZIM index entries for any article whose title or URL contains `|` fails to open the article or fails to follow redirects properly.
- Corrupted directory entry metadata when reconstructed from HTML attributes in UI lists.

## Affected Components

- `www/js/lib/zimDirEntry.js` (`DirEntry.prototype.toStringId`, `DirEntry.fromStringId`)
- `www/js/app.js` (`populateListOfArticles`, `handleTitleClick`, `findDirEntryFromDirEntryIdAndLaunchArticleRead`)
- `www/js/lib/zimArchive.js` (`parseDirEntryId`)

## Acceptance Criteria

1. `DirEntry.prototype.toStringId` and `DirEntry.fromStringId` properly encode/escape delimiter characters in `url` and `title` so round-tripping preserves all fields without corruption.
2. Reconstructed `DirEntry` preserves exact `url`, `title`, `redirect`, and `redirectTarget` values regardless of whether `title` or `url` contains `|`.
3. Selecting search results or ZIM Index items with `|` in titles/URLs correctly resolves redirects and loads the expected article.
4. Automated unit tests verify serialization/deserialization fidelity for entries with pipe characters and special symbols.

## Solution Direction

Encode variable string fields (`url` and `title`) when serializing in `toStringId` (e.g. via `encodeURIComponent`), and decode them in `fromStringId` (via `decodeURIComponent`), or escape `|` characters before joining.

### LINKED ISSUE #1491: DirEntry.toStringId()/fromStringId() corrupts title when url or title contains "|"
## Summary

`DirEntry.prototype.toStringId()` in `www/js/lib/zimDirEntry.js` joins several DirEntry fields with `|` as a separator, and `DirEntry.fromStringId()` splits on `|` to reconstruct them. If `url` or `title` themselves contain a literal `|`, the split produces the wrong field boundaries.

app.js wraps the whole id in `encodeURIComponent()` before writing it to a DOM attribute (`buildArticleListElement`), and reverses that with `decodeURIComponent()` before calling `fromStringId()` (`findDirEntryFromDirEntryIdAndLaunchArticleRead`). That outer encode/decode pair does not protect the inner `|` separators, since it encodes and decodes literal `|` characters the same way it encodes/decodes any other character in the string — so the ambiguity survives the round trip unchanged.

## Real-world impact

I scanned real ZIM archives with `|` in their directory entries (a small node script that reads the ZIM directory listing directly, no cluster decompression needed) rather than relying on a constructed example. Two concrete cases:

- `hsm.stackexchange.com` (History of Science and Mathematics Stack Exchange): titles containing inline LaTeX using `|` for absolute value, e.g. `Did Bolzano conclude that $| \mathbb R | \ne | \mathbb N|$?`
- `tonedear.com` (a Zimit-produced archive): titles in the common `<Page title> | <Site name>` shape, e.g. `Contact | Ear Training`

For both, `url` is serialized before `title` in the joined string, so `url` is unaffected and the article opens normally — this is **not** a navigation or redirect failure. The visible symptom is that `title` gets truncated at the first `|` once split, which shows up wherever the truncated title is used elsewhere in the UI (e.g. the article list, or a "return to..." link).

Round-tripping the real title above through the current `toStringId()` → `encodeURIComponent()` → `decodeURIComponent()` → `fromStringId()` pipeline:

```
real title:          "Contact | Ear Training"
reconstructed title: "Contact "
```

## Suggested fix

`encodeURIComponent()` the `url` and `title` fields individually before joining them with `|` in `toStringId()`, and `decodeURIComponent()` them individually (with a safe fallback if decoding fails, to tolerate any legacy unencoded ids already stored) in `fromStringId()`. This is compatible with the outer `encodeURIComponent()`/`decodeURIComponent()` wrapping in app.js, since encoding the `%` characters produced by the inner encoding is itself reversible.

I have a fix + unit tests ready and will open a PR against this issue. This was originally reported/fixed against kiwix-js-pwa (kiwix/kiwix-js-pwa#966), but per CONTRIBUTING.md, shared app code fixes belong here first.

bhuvan-somisetty:
On-port of kiwix/kiwix-js#1492 (merged upstream, fixes kiwix/kiwix-js#1491). Closes #966.

A literal `|` in a title or url was read as the field separator when `DirEntry.fromStringId()` split the id, so titles came back truncated (e.g. `Contact | Ear Training` in Zimit archives, LaTeX titles in Stack Exchange). `url` and `title` are now percent-encoded in `toStringId()` and decoded in `fromStringId()`. The `www/js/lib/zimDirEntry.js` change is the same as upstream's.

Adds `test/zimDirEntry.test.cjs`, porting the upstream round-trip tests to the node:test runner, plus a CHANGELOG line.

## Testing

- `npm test`: 71 pass. With the `zimDirEntry.js` change stashed, 4 of the 5 new tests fail, so they do catch the bug.
- `npx eslint` clean on both changed files.
- The entries in the tests mirror real ones from hsm.stackexchange.com and a tonedear.com Zimit archive, as verified upstream.
- Manual browser check (source, Vite dev server, Chrome, non-SW mode) with the real `hsm.stackexchange.com_en_all_2026-08.zim`, which has 3 titles containing `|` (found with `scripts/scan-zim-dirents.cjs`): title search for "Did Bolzano conclude" lists `Did Bolzano conclude that $| \mathbb R | \ne | \mathbb N|$?` in full, the same entry appears in the ZIM Archive Index under D, and clicking it opens the correct article. No errors related to the change.
- Caveat: with the fix reverted, the same click-through also worked on this archive, because the entry's url has no `|` and the lookup uses the url. So this check confirms there is no regression rather than reproducing the bug; the bug itself is covered by the unit tests, which fail without the fix.
- Bundled build (`rollup --config --file dist/www/js/bundle.js`, served statically): same result for title search, ZIM Archive Index and opening the article; the bundle contains the encoded `toStringId`.


Jaifroid (changes requested):
Thanks @bhuvan-somisetty, this is a faithful port. The `zimDirEntry.js` change matches upstream exactly, the file's BOM is preserved, and I've confirmed locally that with the fix reverted 4 of your 5 new tests fail, so the suite really does catch the bug. I also checked that nothing else in this repo builds or parses a stringId, and that the libzim search path goes through the same `DirEntry`, so there's nothing PWA-specific left unported.

One change before I merge, in the CHANGELOG line. The search results and ZIM Index lists themselves were never truncated: they render `dirEntry.getTitleOrUrl()` straight from the live dirEntry, so the title is complete there. The truncation only appears *after* an entry is selected, in the "Return to ..." link built from the reconstructed dirEntry (`uiUtil.makeReturnLink()`, called around app.js:6390). As written, the entry points users at the wrong symptom.

Could you reword it to something like:

```
* FIX: The url and title in a directory entry id are now URI-encoded, so a literal "|" is no longer read as the field separator, and titles with "|" are not truncated in the "Return to" link (port of kiwix-js #1492)
```

Your testing note is spot on, by the way, and thank you for being explicit about the caveat: navigation uses the url, so on an archive whose urls contain no "|" the click-through works either way. That's exactly why the unit tests are the right place to pin this down.

Jaifroid:
(And please update the branch.)

bhuvan-somisetty:
Thanks for the careful review and for pointing me at the right symptom. I've reworded the CHANGELOG line with your suggested text, and merged the latest main into the branch (no conflicts, tests still pass). Should be ready for another look.

Jaifroid (approved):
Thanks very much, I'll squash-merge now.

bhuvan-somisetty:
Thanks for the review and for merging, @Jaifroid, and for walking me to the right symptom in the CHANGELOG. I enjoyed working on this and I'm happy to keep helping on the project. If there's another issue you'd like a hand with, in this repo or in kiwix-js, just point me at it.

PR #975, opened 2026-09-19
Escape the text typed in Find in article before building regular expressions (#974)

### LINKED ISSUE #974: Find in article: regex metacharacters in the search text throw a SyntaxError, clear highlights and leave stale counters
The Find in article bar passes the text typed by the user straight into `new RegExp()`, without escaping it. Any search text that is not a valid regular expression (for example `c++`, `(`, `[`, `+1`, `foo)` or `?`) throws an uncaught `SyntaxError`. The existing highlighting is removed but the match counters keep their old values, so the bar looks broken. Text that is a valid regex but not meant as one (`a*`, `a.b`) is silently interpreted as a regex, which gives wrong match counts.

The search text flows like this (`findInArticleInitiate()`, `app.js:591`):
```javascript
localSearch.apply(val);                              // app.js:600
var fullTotal = localSearch.countFullMatches(val);   // app.js:602
```

In `www/js/lib/util.js`, `Hilitor.apply()` (`util.js:569`) calls `addAccents()` and `convertCharStr2jEsc()`. Neither escapes regex metacharacters, and `convertCharStr2jEsc()` only escapes `\`, `'` and `"`. `setRegex()` then builds the pattern:
```javascript
var re = '(' + input + ')';
...
matchRegex = new RegExp(re, 'i');   // util.js:428, throws for invalid input
```
`countFullMatches()` (`util.js:440`) has the same problem:
```javascript
var inputMatcher = new RegExp(input, 'ig');   // util.js:449, input is the raw user text
```

`apply()` calls `this.remove()` before `setRegex()`, so when the RegExp constructor throws, the old highlights are already gone. `countFullMatches()` and the counter update are never reached, so the "Full:" / "Partial:" display keeps the previous search's numbers.

### In the app

1. Open any ZIM article (I used `hsm.stackexchange.com_en_all_2026-08.zim`, the article "Did Bolzano conclude that ...").
2. Click the Find in article button (magnifier in the navbar).
3. Type `Bolzano`. Terms are highlighted and the counters show `Full: 34` / `Partial: 34`.
4. Clear the box and type `c++`, or `(`.

Result: every highlight disappears (0 highlighted elements) but the counters still show `Full: 34` / `Partial: 34`. The browser console logs an uncaught error from `util.js:428`:
```text
Uncaught SyntaxError: Invalid regular expression: /([cç]++)/i: Nothing to repeat
Uncaught SyntaxError: Invalid regular expression: /(()/i: Unterminated group
```

Tested with the source app served by Vite (Chrome, non-Service-Worker mode, version 3.8.93-E).

### In Node

```javascript
import util from './www/js/lib/util.js';

const node = { innerHTML: 'I like c++ and (brackets) and [arrays] and a*b',
    getElementsByClassName: () => [], hasChildNodes: () => false, nodeName: 'BODY', nodeType: 1 };

for (const input of ['c++', '(', '[', '?', 'foo)', '+1']) {
    const h = new util.Hilitor(node);
    h.setMatchType('open');
    try { h.apply(input); } catch (e) { console.log(JSON.stringify(input), 'apply:', e.message); }
    try { h.countFullMatches(input); } catch (e) { console.log(JSON.stringify(input), 'count:', e.message); }
}
```

Output:
```text
"c++" apply: Invalid regular expression: /([cç]++)/i: Nothing to repeat
"c++" count: Invalid regular expression: /c++/gi: Nothing to repeat
"(" apply: Invalid regular expression: /(()/i: Unterminated group
"(" count: Invalid regular expression: /(/gi: Unterminated group
"[" apply: Invalid regular expression: /([)/i: Unterminated character class
"[" count: Invalid regular expression: /[/gi: Unterminated character class
"?" apply: Invalid regular expression: /(?)/i: Invalid group
"foo)" apply: Invalid regular expression: /(f([oôöó]|oe)([oôöó]|oe)))/i: Unmatched ')'
"foo)" count: Invalid regular expression: /foo)/gi: Unmatched ')'
"+1" apply: Invalid regular expression: /(+1)/i: Nothing to repeat
"+1" count: Invalid regular expression: /+1/gi: Nothing to repeat
```

Inputs that happen to be valid regexes are interpreted as patterns instead of literal text: `a*` reports 47 matches in the test string above (`*` is a quantifier, so the pattern matches at almost every position), `f(x)` matches `fx`, `1+1` matches `11`, and `a.b` treats `.` as a wildcard.

## Impact

- Searching for very common terms such as `C++`, `[1]`, `+1` or a lone `(` or `?` breaks the bar: the highlights disappear and the counters show stale numbers, with the error only in the console.
- Other text is silently searched as a pattern instead of literally. For example, `f(x)` matches the text `fx` and not `f(x)` (count 0 for the literal, 1 for the wrong text), and `1+1` matches `11` but not `1+1`.
- The "Full" count can be inflated for text containing `*`, `.` or `|`.
- Some inputs are worse than stale counters. Text whose accidental pattern can match an empty string, such as a lone `^`, makes `hiliteWords()` loop without end: it walks a live `NodeList` while `splitText()` keeps inserting new siblings whose text is never consumed, so the article iframe hangs. (Pointed out by @Jaifroid in the review of #975; I reproduced it in a Node test with a fake DOM.)
- A second crash type comes from input that closes the wrapping group early, for example `a)|(b`. When the second alternative matches, group 1 does not take part in the match and `regs[1]` is `undefined`, so `hiliteWords()` throws `Cannot read properties of undefined (reading 'toLowerCase')` (`util.js:532`). Also reproduced with a fake DOM.

## Affected Components

- `www/js/lib/util.js` (`Hilitor.prototype.apply` / `setRegex`, `countFullMatches`)
- `www/js/app.js` (`findInArticleInitiate`)

## Acceptance Criteria

1. Typing text that contains regular-expression metacharacters (`. * + ? ^ $ { } ( ) | [ ] \`) into Find in article searches for that literal text, highlights it, and updates the counters, without throwing.
2. `countFullMatches()` and the highlighting agree on what counts as a match for the same input.
3. If a pattern still cannot be built, the previous highlights and counters are cleared or left consistent instead of showing stale numbers.
4. Existing behaviour is unchanged for ordinary text: multiple words are still joined with OR, accent-insensitive matching (`addAccents`) still works, and the `\uXXXX` escapes produced by `convertCharStr2jEsc()` still work for non-ASCII text.
5. Unit tests cover inputs such as `c++`, `(`, `[`, `+1`, `?`, `a*` and `a.b`.

## Solution Direction

Escape the user's text before it is turned into a pattern. The escaping has to be applied to the literal characters only, so that the alternation `|` used to join words, the character classes added by `addAccents()` and the `\uXXXX` sequences added by `convertCharStr2jEsc()` are not escaped twice. The simplest place is per word, before `addAccents()`, or by building the pattern from an escaped copy of the input. `countFullMatches()` needs the same escaping. As a safety net, wrapping the `new RegExp()` calls in `try/catch` would keep the UI consistent if anything unexpected slips through.

bhuvan-somisetty:
Fixes the Find in article bar treating the text you type as a regular expression. Closes #974.

`Hilitor.apply()`, `countFullMatches()` and `scrollToFullMatch()` in `www/js/lib/util.js` built a `RegExp` straight from the typed text. Text that is not a valid regex (`c++`, `(`, `[`, `+1`, `foo)`) threw a `SyntaxError`, which cleared the highlights but left the "Full / Partial" counters showing the previous search. Text that happens to be a valid regex was searched for as a pattern: `f(x)` matched `fx` and not `f(x)`, `1+1` matched `11`, and `a*` gave an inflated count.

The typed text is now escaped with a small private `escapeRegExp()` before it is used in any of the three places. In `apply()` this happens before `addAccents()`, so the character classes it adds are not escaped, and the `\uXXXX` escapes from `convertCharStr2jEsc()` still work. A `|` is written as `[|]` instead of `\|`: `setRegex()` joins words with `|` and trims a leading or trailing one, and that would strip the tail of an escaped pipe and leave a lone backslash.

This is the PWA's own feature (kiwix-js has no Find in article), so it belongs in this repo rather than upstream. I have not added a CHANGELOG entry, following CONTRIBUTING; happy to add one if you want it.

Adds `test/hilitor.test.cjs`, which runs the real `Hilitor` source against a minimal fake DOM.

## Testing

- `npm test`: 97 pass (26 of them new). With the `util.js` change stashed, 21 of the 26 new tests fail; the 5 that still pass are the checks that ordinary text (accents, non-ASCII, case, counts) behaves as before. Undoing only the `scrollToFullMatch()` line makes its test fail too.
- `npx eslint test/hilitor.test.cjs` is clean. `util.js` has 27 existing ESLint problems on `main` and the same 27 with this change; none are on the lines I touched.
- Manual check in the source app (Vite dev server, Chrome, non-Service-Worker mode) with `hsm.stackexchange.com_en_all_2026-08.zim`: in Find in article, `Bolzano` gives 34 highlights; then `$|`, `(` and `c++` each update the counters and highlights (`(` gives 17) with no console errors. Before the change, `c++` and `(` cleared the highlights, left the counters at 34 and logged `Uncaught SyntaxError: Invalid regular expression` from `util.js`. While testing I found that `scrollToFullMatch()` had the same problem, which is why it is included here.
- Not tested: the bundled build, other browsers, and the Electron and NWJS apps.


Jaifroid (approved):
Thanks @bhuvan-somisetty -- I've checked this over and I'm happy to merge it.

I reproduced the issue on `main` and verified the fix independently:

- **Coverage**: all three `RegExp`-building sites in `Hilitor` are patched. `getRegex()` is the only other regex consumer and has no caller outside `util.js`, so nothing is left.
- **Robustness**: fuzzing 200,000 random terms over letters plus punctuation gives 79,945 failures on `main` and 0 on this branch.
- **No regression**: 50,000 fuzzed terms containing no metacharacters, plus a set of real-world terms covering accents, ß/œ, apostrophes, hyphens, CJK, multi-word ('left' match type) and case, give byte-identical results before and after.
- **Ordering**: escaping before `addAccents()` is right, and non-obvious: `convertCharStr2jEsc()` doubles the backslash and `setRegex()`'s `replace(/\\([^u]|$)/g, '$1')` un-doubles it, so the escape round-trips exactly. Worth the comment you left on it.
- **The `[|]` case**: also necessary rather than merely defensive: `setRegex()` strips the backslash first, and the trim then removes the bare `|`, so a search for a pipe would silently return nothing. Good catch, and relevant to ZIMs with LaTeX in their titles.
- **Bundled build** (your untested item): `npm run build-min` succeeds and the helper survives minification intact, so that gap is closed.

Two symptoms worth adding to #974, as they are worse than the stale counters:

1. Terms whose accidental pattern can match an empty string hang the article iframe rather than just breaking the counters: `hiliteWords()` iterates a live `NodeList` at `util.js:526` while `splitText()` keeps inserting siblings whose text is never consumed.
2. There is a second crash type, `Cannot read properties of undefined (reading 'toLowerCase')` at `util.js:532`, when a non-participating group leaves `regs[1]` undefined.

Both are fixed by this PR.

Minor and not blocking: `countFullMatches()` still collapses `.,;:?!¿¡-` to a space before escaping, so a term of pure punctuation counts spaces rather than matches. It is masked by the `fullTotal`/`partialTotal` clamp in `app.js` and is pre-existing, so please leave it as is.

Could you add the following CHANGELOG text? :-

```
* FIX: Bug that failed to handle regex characters typed in Find in article, causing stale match counters and hangs
```

When you've done that, let me know, and I'll squash/merge.

bhuvan-somisetty:
Thanks for the thorough review and the fuzzing, @Jaifroid. I've added the CHANGELOG line exactly as you wrote it, and merged the latest main into the branch (tests and CI are green). I also added the hang and the `toLowerCase` crash to #974 with credit to you; I could reproduce both before the fix (a lone `^` and `a)|(b`) and neither happens now. Ready for you to squash and merge, and I'll leave the punctuation counting in `countFullMatches()` as it is.

